PR #1, opened 2022-04-30
Solution

liuda-chuzdiuk:


serhii56465 (approved):

cinema/serializers.py
```diff
@@ -0,0 +1,23 @@
...
+from cinema.models import Movie
+
+
+class MovieSerializer(serializers.Serializer):
+    id = serializers.IntegerField(read_only=True)
+    title = serializers.CharField(
+        max_length=255,
+        required=True
+    )
+    description = serializers.CharField(max_length=255, required=False)
```
IvanRamyk, line 12: if the description is not required it'd be better to set `blank=True, null=True` in its model

cinema/serializers.py
```diff
@@ -0,0 +1,23 @@
...
+    duration = serializers.IntegerField(required=False)
```
IvanRamyk, line 13: if the duration is not required it'd be better to set `blank=True, null=True` in its model

cinema_service/urls.py
```diff
@@ -0,0 +1,23 @@
+"""cinema_service URL Configuration
```
IvanRamyk, line 1: Please, remove these comments

IvanRamyk (changes requested):

IvanRamyk (approved):
Well done!

PR #4, opened 2022-05-01
Solution

denyslipin:


IvanRamyk (changes requested):
Remove __pycache__ from PR

app/urls.py
```diff
@@ -0,0 +1,22 @@
+"""app URL Configuration
```
IvanRamyk, line 1: remove this comments, please

cinema/serializers.py
```diff
@@ -0,0 +1,22 @@
+from rest_framework import serializers
+
+from cinema.models import Movie
+
+
+class MovieSerializer(serializers.Serializer):
+    id = serializers.IntegerField(read_only=True)
+    title = serializers.CharField(required=True)
+    description = serializers.CharField(required=False)
```
IvanRamyk, line 9: if the description is not required it'd be better to set `blank=True, null=True` in its model

IvanRamyk, line 9: if the description is not required it'd be better to set `blank=True, null=True` in its model

IvanRamyk (changes requested):

IvanRamyk (approved):
Good job!

PR #11, opened 2022-05-04
Solution

Geksus:


cinema/views.py
```diff
@@ -0,0 +1,39 @@
...
+    if request.method == "POST":
+        serializer = MovieSerializer(data=request.data)
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_201_CREATED)
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+
+
+@api_view(["GET", "PUT", "DELETE"])
+def movie_update_delete(request, pk):
```
liuda-chuzdiuk, line 25: You forgot one method,  didn't you?)

liuda-chuzdiuk (changes requested):
.idea, db should not be in PR,  and I am not sure about inspectionProfiles

cinema/views.py
Geksus: I did. Fixed

PR #17, opened 2022-07-12
Solution

Serhii-Leonenko:


cinema/serializers.py
```diff
@@ -0,0 +1,26 @@
+from rest_framework import serializers
+
+from cinema.models import Movie
+
+
+class MovieSerializer(serializers.Serializer):
+    id = serializers.IntegerField(read_only=True)
+    title = serializers.CharField(max_length=55)
+    description = serializers.CharField(required=False, max_length=500)
```
mens0, line 9: i think required=False is set by default

mens0 (approved):

cinema/serializers.py
Serhii-Leonenko: No, on the contrary)

cinema/serializers.py
```diff
@@ -0,0 +1,26 @@
+from rest_framework import serializers
+
+from cinema.models import Movie
+
+
+class MovieSerializer(serializers.Serializer):
+    id = serializers.IntegerField(read_only=True)
+    title = serializers.CharField(max_length=55)
```
TPolina, line 8: why do you have different max_length values in the model and serializer?

TPolina (changes requested):

cinema/serializers.py
Serhii-Leonenko: yes, my mistake

TPolina (approved):

PR #21, opened 2022-07-12
Develop

KatiukhaO:


cinema/models.py
```diff
@@ -0,0 +1,13 @@
+from django.db import models
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255, blank=False, null=False)
+    description = models.TextField(blank=True)
+    duration = models.TimeField(blank=True, null=True)
+
+    class Meta:
+        verbose_name_plural = "movies"
```
katevasilenko, line 10: I'm not sure you really need it, because django creates plural right by itself, but okey

katevasilenko (approved):

cinema/models.py
```diff
@@ -0,0 +1,13 @@
+from django.db import models
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255, blank=False, null=False)
```
TPolina, line 5: blank=False, null=False are default values, no need to write it

cinema/models.py
```diff
@@ -0,0 +1,13 @@
...
+    description = models.TextField(blank=True)
+    duration = models.TimeField(blank=True, null=True)
```
TPolina, line 7: duration is not a TimeField and also it shouldn't be blank or null

cinema/models.py
```diff
@@ -0,0 +1,13 @@
+from django.db import models
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255, blank=False, null=False)
+    description = models.TextField(blank=True)
```
TPolina, line 6: will be better to set null=True as well

cinema/serializer.py
```diff
@@ -0,0 +1,22 @@
+from rest_framework import serializers
+
+from cinema.models import Movie
+
+
+class MovieSerializer(serializers.Serializer):
+    id = serializers.IntegerField(read_only=True)
+    title = serializers.CharField(max_length=255, required=False)
```
TPolina, line 8: title and duration must be required

cinema/serializer.py
```diff
@@ -0,0 +1,22 @@
...
+    description = serializers.CharField(required=False)
+    duration = serializers.TimeField(required=False)
```
TPolina, line 10: not a TimeField

TPolina (changes requested):

TPolina (approved):

PR #22, opened 2022-07-12
Katya's Solution

katevasilenko:


cinema/views.py
```diff
@@ -0,0 +1,45 @@
...
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_201_CREATED)
+
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+
+
+@api_view(["GET", "PUT", "DELETE"])
+def movie_detail(request, pk):
+    movie = Movie.objects.get(id=pk)
```
DmytroSvirsa, line 28: There is recomendation from django documentation to handle not valid ids:
````
try:
    movie = Movie.objects.get(pk=pk)
except Movie.DoesNotExist:
    return Response(status=status.HTTP_404_NOT_FOUND)
````

DmytroSvirsa (approved):

cinema/views.py
TPolina: Dima is right you have to handle non existing pk somehow

TPolina (changes requested):

PR #23, opened 2022-07-12
movie_api

CheshkoSergey:


NOSSUDAK (approved):

cinema/models.py
```diff
@@ -0,0 +1,10 @@
+from django.db import models
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=100, unique=True, blank=False)
```
TPolina, line 5: blank=False is a default value, no need to write it

TPolina (changes requested):

TPolina (approved):

PR #28, opened 2022-07-12
Solution

TsybuliaRoman:


cinema/models.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.CharField(max_length=255, null=True, blank=True)
+    duration = models.IntegerField()
+
+    class Meta:
+        verbose_name = "movie"
+        verbose_name_plural = "movies"
```
katevasilenko, line 11: I'm not sure you really need it, because django do it well by itself, but, ok, why not

katevasilenko (approved):

PR #35, opened 2022-07-12
Solution

vasylhnatiuk:


cinema/serializers.py
```diff
@@ -0,0 +1,19 @@
+from rest_framework import serializers
+from cinema.models import Movie
+
+
+class MovieSerializer(serializers.Serializer):
+    id = serializers.IntegerField(read_only=True)
+    title = serializers.CharField(required=False)
```
katevasilenko, line 7: and here required=True

cinema/urls.py
```diff
@@ -0,0 +1,11 @@
+from django.urls import path
+
+from cinema.views import movie_list, change_movie
+
+urlpatterns = [
+    path("movies/", movie_list, name="movie-list"),
+    path("movies/<int:pk>/", change_movie, name="change-movie"),
```
katevasilenko, line 7: maybe it's not important, but you also can read movie in this url, so it's better to write movie_detail

cinema/views.py
```diff
@@ -0,0 +1,43 @@
...
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_201_CREATED)
+
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+
+
+@api_view(["GET", "PUT", "DELETE"])
+def change_movie(request, pk):
+    movie = Movie.objects.get(id=pk)
```
katevasilenko, line 28: I made this mistake too) You should check if it pk exists:
`

> try:
>     movie = Movie.objects.get(pk=pk)
> except Movie.DoesNotExist:
>     return Response(status=status.HTTP_404_NOT_FOUND)

`

MasterpieceElbow (changes requested):
You have fixed `.gitignore`. But you still have `.idea` files in this PR.

MasterpieceElbow (approved):

PR #37, opened 2022-07-12
Solution

Evgeniy994:


daria-niklaevna (approved):

cinema/models.py
```diff
@@ -0,0 +1,10 @@
+from django.db import models
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField(blank=True)
```
TPolina, line 6: better set null=True as well

cinema/urls.py
```diff
@@ -0,0 +1,10 @@
+from django.urls import path
+
+from cinema.views import movie_list, movie_detail
+
+urlpatterns = [
+    path("movies/", movie_list, name="movie-list"),
+    path("movies/<int:pk>/", movie_detail, name="movie-detail"),
+]
+
+app_name = "cinema"
```
TPolina, line 10: flake8

TPolina (changes requested):

MasterpieceElbow (approved):

PR #75, opened 2022-11-08
Implemented an API that provides CRUD operations

AlenOl:
Created a Django project and a cinema app inside it. Implemented a serializer for this model.


DmytroSvirsa (changes requested):
remove .idea files and db.sqlite3

DmytroSvirsa (changes requested):
In general god job, but you still have db.sqlite3 here, ant it will be a god practise to add .gitignore file to not push garbage files

DmytroSvirsa (approved):

PR #77, opened 2022-11-13
Solution

KirillDrago:


shurushku (changes requested):
There is something wrong with your `.gitignore`. You have `.idea` and `db.sqlite3` in your solution.
I think this should be fixed.

TPolina (changes requested):
You still have `.idea` in your pr

TPolina (approved):
Well done!

PR #80, opened 2022-11-14
Solution

StasOkhrym:


cinema/serializers.py
```diff
@@ -0,0 +1,21 @@
+from rest_framework import serializers
+
+from cinema.models import Movie
+
+
+class MovieSerializer(serializers.Serializer):
+    id = serializers.IntegerField(read_only=True)
+    title = serializers.CharField(required=True, max_length=255)
```
rostyak, line 8: I think we don't need to write here parameters like max_length, etc. At least it works without them) But maybe it's good practice, idk

rostyak (approved):
Good

TPolina (approved):
Well done!

PR #84, opened 2022-11-14
Solution

Vasyl-Poremchuk:


cinema/serializers.py
```diff
@@ -0,0 +1,27 @@
+from rest_framework import serializers
+
+from cinema.models import Movie
+
+
+class MovieSerializer(serializers.Serializer):
+    id = serializers.IntegerField(read_only=True)
+    title = serializers.CharField(required=True)
```
linaSam, line 8: "required=True" is default, no need to write it

cinema/views.py
```diff
@@ -1,3 +1,56 @@
...
+
+            return Response(serializer.data, status=status.HTTP_201_CREATED)
+
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+
+
+@api_view(["GET", "PUT", "DELETE"])
+def movie_detail(request, pk):
+    """Retrieve, update or delete a code movie."""
+    try:
```
linaSam, line 33: you can use "get_object_or_404"

linaSam (changes requested):

cinema/models.py
```diff
@@ -0,0 +1,7 @@
+from django.db import models
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
```
wQuelS, line 6: I guess you can set blank=True for the description field. But it seems like it is not an argument for reject:)

cinema/views.py
```diff
@@ -0,0 +1,56 @@
...
+            serializer.save()
+
+            return Response(serializer.data, status=status.HTTP_201_CREATED)
+
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+
+
+@api_view(["GET", "PUT", "DELETE"])
+def movie_detail(request, pk):
+    """Retrieve, update or delete a code movie."""
```
wQuelS, line 32: I like that you writing doc strings, but you have to make all of them in one style. Here DS doesn't have a blank line after it, as it has in the previous function. This change is up to you, btw.

wQuelS (approved):

PR #86, opened 2022-11-15
Solution!

ketstap162:


cinema/views.py
```diff
@@ -0,0 +1,46 @@
...
+
+        if serializer.is_valid():
+            serializer.update(movie, validated_data=request.data)
+            return Response(serializer.data, status=status.HTTP_200_OK)
+
+        return Response(serializer.errors, status=status.HTTP_200_OK)
+
+    if request.method == "DELETE":
+        Movie.objects.filter(pk=pk).delete()
+        return Response(status=status.HTTP_200_OK)
```
Vasyl-Poremchuk, line 46: I think it will be `status=status.HTTP_204_NO_CONTENT`

cinema/views.py
```diff
@@ -0,0 +1,46 @@
...
+
+    if request.method == "PUT":
+        movie = Movie.objects.get(pk=pk)
+        serializer = MovieSerializer(data=request.data)
+
+        if serializer.is_valid():
+            serializer.update(movie, validated_data=request.data)
+            return Response(serializer.data, status=status.HTTP_200_OK)
+
+        return Response(serializer.errors, status=status.HTTP_200_OK)
```
Vasyl-Poremchuk, line 42: I think it will be `status=status.HTTP_400_BAD_REQUEST`

Vasyl-Poremchuk (changes requested):

Vasyl-Poremchuk (approved):
GJ!

cinema/models.py
```diff
@@ -0,0 +1,10 @@
+from django.db import models
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.CharField(max_length=255, null=True)
```
TPolina, line 6: for description it's better to write null=True, blank=True

cinema/views.py
```diff
@@ -0,0 +1,46 @@
+from django.shortcuts import render
```
TPolina, line 1: do you use this import?

cinema/views.py
```diff
@@ -0,0 +1,46 @@
...
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_201_CREATED)
+
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+
+
+@api_view(["GET", "PUT", "DELETE"])
+def movie_pk(request, pk):
+    if request.method == "GET":
+        movie = Movie.objects.get(pk=pk)
```
TPolina, line 30: You need to catch an error in case the object doesn't exist. Also, you can take this outside if statements not to repeat your code

cinema/views.py
```diff
@@ -0,0 +1,46 @@
...
+        serializer = MovieSerializer(movie)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+
+    if request.method == "PUT":
+        movie = Movie.objects.get(pk=pk)
+        serializer = MovieSerializer(data=request.data)
```
TPolina, line 36: you need to pass movie as well

cinema/views.py
```diff
@@ -0,0 +1,46 @@
...
+
+        if serializer.is_valid():
+            serializer.update(movie, validated_data=request.data)
```
TPolina, line 39: why not just use serializer.save()?

TPolina (changes requested):

cinema/views.py
ketstap162: Because we need update existing movie, not create new

cinema/views.py
ketstap162: ?

cinema/serializers.py
```diff
@@ -0,0 +1,20 @@
+from rest_framework import serializers
+
+from cinema.models import Movie
+
+
+class MovieSerializer(serializers.Serializer):
+    id = serializers.IntegerField(read_only=True)
+    title = serializers.CharField(required=True)
```
Y-Havryliv, line 8: https://www.django-rest-framework.org/api-guide/fields/#required
it's the default value

cinema/urls.py
```diff
@@ -0,0 +1,10 @@
+from django.urls import path
+
+from cinema.views import movie_list, movie_pk
+
+urlpatterns = [
+    path("movies/", movie_list, name="cinema-list"),
```
Y-Havryliv, line 6: strange names, `movie_list` and `movie_detail` would be better

cinema/views.py
Y-Havryliv: `MovieSerializer(movie, data=request.data)`

Nattalli (approved):

PR #88, opened 2022-11-15
Solution

OvDima1:


cinema/views.py
```diff
@@ -0,0 +1,48 @@
...
+
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_201_CREATED)
+
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+
+
+@api_view(["GET", "PUT", "DELETE"])
+def movie_detailed(request, pk):
```
rostyak, line 27: is movie_detailed a possible name? (not movie_detail?)

rostyak (approved):

cinema/views.py
OvDima1: i think it`s not a problem because we use it in urls.py. Maybe it`s not a best prcatice. Let`s see mentor`s opinion

cinema/urls.py
```diff
@@ -0,0 +1,10 @@
+from django.urls import path
+
+from cinema.views import movie_list, movie_detailed
+
+urlpatterns = [
+    path("movies/", movie_list, name="movie-list"),
+    path("movies/<pk>", movie_detailed, name="movie-detailed")
```
TPolina, line 7: always add `/` to the end of endpoint

cinema/views.py
```diff
@@ -0,0 +1,48 @@
...
+    if request.method == "GET":
+        serializer = MovieSerializer(movie)
+        return Response(serializer.data)
+
+    if request.method == "PUT":
+        serializer = MovieSerializer(movie, data=request.data)
+
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data)
```
TPolina, line 42: status?

TPolina (changes requested):

TPolina (approved):
Good job!

PR #89, opened 2022-11-15
Solution

linaSam:


cinema/views.py
```diff
@@ -0,0 +1,48 @@
...
+
+    if request.method == "POST":
+        serializer = MovieSerializer(data=request.data)
+
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_201_CREATED)
+
+
+@csrf_exempt
```
Vasyl-Poremchuk, line 27: I don't know if you need to use the `@csrf_exempt` decorator here, but I think if you implement it here, you should also apply it in the view function above.
By the way, if we use the `@api_view` decorator, then there is no need to protect against `csrf-attack` because our decorator wraps `APIView` and implements it in the `as_view` method.
But maybe I'm wrong. I don't think you need to change it now, I think the mentor will give you better explanations.

Vasyl-Poremchuk: image

Vasyl-Poremchuk (changes requested):

cinema/views.py
linaSam: thanks! I decided to just use api_view here:)

Vasyl-Poremchuk (approved):
GJ!

.gitignore
```diff
@@ -0,0 +1,344 @@
...
+
+# Cython debug symbols
+
+# PyCharm
+#  JetBrains specific template is maintained in a separate JetBrains.gitignore that can
+#  be found at https://github.com/github/gitignore/blob/main/Global/JetBrains.gitignore
+#  and can be added to the global gitignore or merged into this file.  For a more nuclear
+#  option (not recommended) you can uncomment the following to ignore the entire idea folder.
+
+# End of https://www.toptal.com/developers/gitignore/api/python,pycharm+all,django
```
Y-Havryliv, line 344: add blank line

cinema/views.py
```diff
@@ -0,0 +1,49 @@
...
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_201_CREATED)
+
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+
+
+@api_view(["GET", "PUT", "DELETE"])
+def movie_detail(request, pk):
+    try:
```
Y-Havryliv, line 31: use `get_object_or_404`

cinema/views.py
```diff
@@ -0,0 +1,49 @@
...
+        snippet = Movie.objects.get(pk=pk)
+    except Movie.DoesNotExist:
+        return HttpResponse(status=status.HTTP_404_NOT_FOUND)
+
+    if request.method == 'GET':
+        serializer = MovieSerializer(snippet)
+        return Response(serializer.data)
```
Y-Havryliv, line 38: add `status code`

cinema/views.py
```diff
@@ -0,0 +1,49 @@
...
+
+    elif request.method == 'PUT':
+        serializer = MovieSerializer(snippet, data=request.data)
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data)
```
Y-Havryliv, line 44: add `status code`

Y-Havryliv (changes requested):

cinema/models.py
```diff
@@ -0,0 +1,13 @@
+from django.db import models
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.CharField(max_length=255)
+    duration = models.IntegerField()
+
+    class Meta:
+        verbose_name_plural = "movies"
```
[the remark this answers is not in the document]
linaSam: probably on the principle "why not" or that it can be useful in future... somehow

Y-Havryliv (approved):
GJ!

PR #90, opened 2022-11-15
Develop

Aleksandro777:


cinema/urls.py
```diff
@@ -0,0 +1,10 @@
+from django.urls import path
+
+from .views import movie_list, movie_detail
+
+urlpatterns = [
+    path("movie/", movie_list, name="movie-list"),
```
rostyak, line 6: I think we need to write our endpoint in the plural ("movies/", "movies/<int:pk>/")

cinema/views.py
```diff
@@ -0,0 +1,45 @@
...
+
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_200_OK)
+
+        return Response(serializer.errors, status=400)
+
+    if request.method == "DELETE":
+        movie.delete()
+        return Response(status=status.HTTP_200_OK)
```
rostyak, line 45: shouldn't the status be 204 here?

rostyak (changes requested):

cinema/urls.py
Aleksandro777: Done

cinema/views.py
```diff
@@ -0,0 +1,45 @@
...
+        return Response(serializer.data, status=status.HTTP_200_OK)
+
+    if request.method == "PUT":
+        serializer = MovieSerializer(movie, data=request.data)
+
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_200_OK)
+
+        return Response(serializer.errors, status=400)
```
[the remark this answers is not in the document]
Aleksandro777: Done

cinema/views.py
Aleksandro777: image
stack overflow says that we can use both.

cinema/urls.py
```diff
@@ -0,0 +1,10 @@
+from django.urls import path
+
+from .views import movie_list, movie_detail
+
+urlpatterns = [
+    path("movies/", movie_list, name="movie-list"),
+    path("movies/<int:pk>/", movie_detail, name="movie_detail"),
```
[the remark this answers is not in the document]
Aleksandro777: ok, thanks

PR #92, opened 2022-11-15
Solution

LiudmylaKulinchenko:


shurushku (approved):

cinema/views.py
```diff
@@ -0,0 +1,55 @@
+from django.db import transaction
+from django.shortcuts import render
```
TPolina, line 2: do you use these imports?

TPolina (changes requested):

TPolina (changes requested):
Delete `.idea` from Github

TPolina (approved):
Good job!

PR #93, opened 2022-11-15
Solution

Kitazuka:


cinema/serializers.py
```diff
@@ -0,0 +1,20 @@
+from rest_framework import serializers
+
+from cinema.models import Movie
+
+
+class MovieSerializer(serializers.Serializer):
+    id = serializers.IntegerField(read_only=True)
+    title = serializers.CharField(max_length=255)
```
IvanStored, line 8: I think, movie can`t exist without title :)

IvanStored (changes requested):

IvanStored (approved):

cinema/serializers.py
```diff
@@ -0,0 +1,20 @@
+from rest_framework import serializers
+
+from cinema.models import Movie
+
+
+class MovieSerializer(serializers.Serializer):
+    id = serializers.IntegerField(read_only=True)
+    title = serializers.CharField(max_length=255, required=True)
+    description = serializers.CharField(max_length=255)
+    duration = serializers.IntegerField(required=True)
```
TPolina, line 10: required=True is a default value. False is not. If you want to make description not required you need to specify this in model as well. If you want to make description 255 char max you need to use Charfield for this field in model

cinema/serializers.py
```diff
@@ -0,0 +1,20 @@
...
+
+    def create(self, validated_data) -> Movie:
+        return Movie.objects.create(**validated_data)
+
+    def update(self, instance, validated_data) -> Movie:
+        instance.title = validated_data.get("title", instance.title)
+        instance.description = validated_data.get("title", instance.description)
+        instance.duration = validated_data.get("title", instance.duration)
```
TPolina, line 18: why `title` is everywhere?

cinema/views.py
```diff
@@ -0,0 +1,42 @@
+from rest_framework import status
+from rest_framework.decorators import api_view
+from rest_framework.response import Response
+
+from cinema.models import Movie
+from cinema.serializers import MovieSerializer
+
+
+@api_view(["GET", "POST"])
+def movies_list(request):
```
TPolina, line 10: movie-list is better

cinema/views.py
```diff
@@ -0,0 +1,42 @@
...
+        serializer = MovieSerializer(data=request.data)
+
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_201_CREATED)
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+
+
+@api_view(["GET", "PUT", "DELETE"])
+def movies_update(request, pk):
```
TPolina, line 26: movie-detail is better

cinema/views.py
```diff
@@ -0,0 +1,42 @@
...
+    except Movie.DoesNotExist:
+        return Response(status=status.HTTP_404_NOT_FOUND)
+
+    if request.method == "GET":
+        serializer = MovieSerializer(movie)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+
+    if request.method == "PUT":
+        serializer = MovieSerializer(movie, data=request.data)
+        return Response(serializer.data, status=status.HTTP_200_OK)
```
TPolina, line 38: Where is the check that serializer is valid?

TPolina (changes requested):

cinema/urls.py
```diff
@@ -1,10 +1,10 @@
 from django.urls import path
 
-from cinema.views import movies_list, movies_update
+from cinema.views import movie_list, movie_update
 
 urlpatterns = [
-    path("movies/", movies_list, name="movies-list"),
-    path("movies/<int:pk>/", movies_update, name="movies-update"),
+    path("movies/", movie_list, name="movies-list"),
+    path("movies/<int:pk>/", movie_update, name="movies-update"),
```
Nattalli, line 7: btw, this is not only about updating. This is also about getting and deleting

Nattalli (approved):

PR #100, opened 2022-11-15
Solution

Mowka7:
I accidentally mixed up the name of an app so... I've just changed the name in json file. It works, so... next time I will pay more attention

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -3,6 +3,6 @@ from django.db import models
 
 class Movie(models.Model):
     title = models.CharField(max_length=255)
-    description = models.TextField(null=True)
+    description = models.TextField(null=True, blank=True)
     duration = models.IntegerField()
 
--- a/app/urls.py
+++ b/app/urls.py
@@ -4,7 +4,7 @@ from app.views import movies_list_view, movie_detail_view
 
 urlpatterns = [
     path('movies/', movies_list_view, name="movie-list"),
-    path('movies/<int:pk>', movie_detail_view, name="movie-detail")
+    path('movies/<int:pk>/', movie_detail_view, name="movie-detail")
 ]
 
 app_name = 'app'


cinema_serviсe_db_data.json
```diff
@@ -1,6 +1,6 @@
 [
```
Aleksandro777, line 1: I think u should remove this file from Git Hub

app/serializers.py
```diff
@@ -0,0 +1,20 @@
...
+
+    def create(self, validated_data):
+        return Movie.objects.create(**validated_data)
+
+    def update(self, instance, validated_data):
+        instance.title = validated_data.get("title", instance.title)
+        instance.description = validated_data.get("description", instance.description)
+        instance.duration = validated_data.get("duration", instance.duration)
+        instance.save()
+        return validated_data
```
Aleksandro777, line 20: I think we should return instance here

Aleksandro777 (changes requested):

Aleksandro777 (approved):
Super!

app/models.py
```diff
@@ -0,0 +1,8 @@
+from django.db import models
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField(null=True)
```
TPolina, line 6: for description it's better to write null=True, blank=True

app/urls.py
```diff
@@ -0,0 +1,10 @@
+from django.urls import path
+
+from app.views import movies_list_view, movie_detail_view
+
+urlpatterns = [
+    path('movies/', movies_list_view, name="movie-list"),
+    path('movies/<int:pk>', movie_detail_view, name="movie-detail")
```
TPolina, line 7: always add `/` to the end of endpoint

TPolina (changes requested):

PR #103, opened 2022-11-15
Develop

alexzubkoff:


.gitignore
```diff
@@ -0,0 +1,137 @@
...
+Control.system-ca-bundle
+GitHub.sublime-settings
+
+# Visual Studio Code #
+.vscode/*
+!.vscode/settings.json
+!.vscode/tasks.json
+!.vscode/launch.json
+!.vscode/extensions.json
+.history
```
mltdll, line 137: It's better to always end text files with a newline character. You can read more about why this is the case [here](https://stackoverflow.com/a/729795).

cinema/views.py
```diff
@@ -0,0 +1,45 @@
+from django.http import HttpResponse, JsonResponse
+from django.views.decorators.csrf import csrf_exempt
+from rest_framework.parsers import JSONParser
+from cinema.models import Movie
+from cinema.serializers import MovieSerializer
+
+
+@csrf_exempt
```
mltdll, line 8: You can use rest_framework's `@api_view` instead.

cinema/views.py
```diff
@@ -0,0 +1,45 @@
...
+def movie_list(request: HttpResponse) -> JsonResponse:
```
mltdll, line 9: In Django, requests are actually of `HttpRequest` type. And since we're using rest framework here, you can import and use `Request` and `Response` here, which would be the correct type annotation.

cinema/views.py
```diff
@@ -0,0 +1,45 @@
...
+        serializer = MovieSerializer(movies, many=True)
+        return JsonResponse(serializer.data, safe=False)
+
+    elif request.method == 'POST':
+        data = JSONParser().parse(request)
+        serializer = MovieSerializer(data=data)
+        if serializer.is_valid():
+            serializer.save()
+            return JsonResponse(serializer.data, status=201)
+        return JsonResponse(serializer.errors, status=400)
```
mltdll, line 21: I'd advise using DRF's `Response` to generate responses here. That way you won't have to use `safe=False`.

cinema/views.py
```diff
@@ -0,0 +1,45 @@
...
+        data = JSONParser().parse(request)
+        serializer = MovieSerializer(Movie, data=data)
+        if serializer.is_valid():
+            serializer.save()
+            return JsonResponse(serializer.data)
+        return JsonResponse(serializer.errors, status=400)
+
+    elif request.method == 'DELETE':
+        Movie.delete()
+        return HttpResponse(status=204)
```
mltdll, line 45: All of the above applies to this function as well.

mltdll (changes requested):

cinema/views.py
alexzubkoff: Thanks, good advice!

cinema/views.py
alexzubkoff: Thanks!

cinema/views.py
alexzubkoff: Ok, thanks! It was science experiment.

mltdll (approved):

cinema/views.py
```diff
@@ -0,0 +1,47 @@
...
+from rest_framework.request import Request
+from rest_framework.response import Response
+
+from cinema.models import Movie
+from cinema.serializers import MovieSerializer
+
+
+@csrf_exempt
+def movie_list(request: Request) -> Response:
+    if request.method == 'GET':
```
Nattalli, line 12: hmm, use a `""` instead of `''`

cinema/views.py
```diff
@@ -0,0 +1,47 @@
...
+        movies = Movie.objects.all()
+        serializer = MovieSerializer(movies, many=True)
+        return Response(serializer.data)
+
+    elif request.method == 'POST':
+        data = JSONParser().parse(request)
```
Nattalli, line 18: can you do it more easy?

Nattalli (changes requested):

PR #104, opened 2022-11-15
Solution

mltdll:


cinema/urls.py
```diff
@@ -0,0 +1,10 @@
+from django.urls import path
+
+from cinema.views import movie_list, movie_detail
+
+app_name = "cinema"
+
+urlpatterns = [
+    path("movies/", movie_list, name="movie-list"),
+    path("movies/<int:pk>/", movie_detail, name="movie_detail"),
```
Ilyakson, line 9: I think you need to replace _ with -

Ilyakson (changes requested):

Ilyakson (approved):
Good!

cinema/views.py
```diff
@@ -0,0 +1,51 @@
...
+
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+
+
+@api_view(["GET", "PUT", "DELETE"])
+def movie_detail(request: Request, pk: int) -> Response:
+    try:
+        movie = Movie.objects.get(pk=pk)
+    except Movie.DoesNotExist:
+        return Response(status=status.HTTP_404_NOT_FOUND)
```
Nattalli, line 32: that's better to use `get_object_or_404` here, even if the difference not a huge

Nattalli (approved):

PR #105, opened 2022-11-16
Solution

VadymShkarbul:


cinema/models.py
```diff
@@ -0,0 +1,10 @@
+from django.db import models
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.CharField(max_length=255)
```
HStakhiv, line 6: Probably better to use `TextField`

cinema/serializers.py
```diff
@@ -0,0 +1,20 @@
+from rest_framework import serializers
+
+from cinema.models import Movie
+
+
+class MovieSerializer(serializers.Serializer):
+    id = serializers.IntegerField(read_only=True)
+    title = serializers.CharField()
+    description = serializers.CharField()
+    duration = serializers.IntegerField()
```
HStakhiv, line 10: Add `required=`

HStakhiv (changes requested):
Ok, but

cinema/models.py
VadymShkarbul: Thanks, I agree!

cinema/serializers.py
VadymShkarbul: Yes, thank you I'll fix it

HStakhiv (approved):
Good

cinema/models.py
```diff
@@ -0,0 +1,10 @@
+from django.db import models
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField(max_length=255)
```
TPolina, line 6: if you want to specify `max_length` it's more reasonable to use CharField here

cinema/serializers.py
```diff
@@ -0,0 +1,26 @@
+from rest_framework import serializers
+
+from cinema.models import Movie
+
+
+class MovieSerializer(serializers.Serializer):
+    id = serializers.IntegerField(read_only=True)
+    title = serializers.CharField(required=False)
+    description = serializers.CharField(max_length=255)
+    duration = serializers.IntegerField(required=False)
```
TPolina, line 10: title, duration can't be required=False

TPolina (changes requested):

cinema/models.py
VadymShkarbul: Understood, thank you!

TPolina (approved):
Good job!

PR #106, opened 2022-11-16
Solution

Yaroslav-Stetsko:


cinema/views.py
```diff
@@ -0,0 +1,47 @@
...
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_201_CREATED)
+
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+
+
+@api_view(["GET", "PUT", "DELETE"])
+def movie_info(request, pk):
+    movie = get_object_or_404(Movie, pk=pk)
+
```
skrypka-dmytro, line 30: read this documentation it will help to make your code better https://www.django-rest-framework.org/tutorial/1-serialization/
and add an annotation type

skrypka-dmytro (changes requested):

cinema/views.py
Yaroslav-Stetsko: I would like some context regarding those improvements...

Yaroslav-Stetsko: This documentation is on serialization we need the requests-and-responses one

Yaroslav-Stetsko: annotation type:
We were told that its is not required

skrypka-dmytro: image

Yaroslav-Stetsko: get_object_or_404 is the same as the method listed in docs so i disagree

skrypka-dmytro: > get_object_or_404 is the same as the method listed in docs so i disagree

Good

cinema/serializers.py
```diff
@@ -0,0 +1,22 @@
...
+    id = serializers.IntegerField(read_only=True)
+    title = serializers.CharField(required=True)
+    description = serializers.CharField(required=True)
+    duration = serializers.IntegerField(required=True)
+
+    def create(self, validated_data):
+        return Movie.objects.create(**validated_data)
+
+    def update(self, instance, validated_data):
+        instance.title = validated_data.get("info", instance.title)
```
TPolina, line 16: info?

cinema/views.py
```diff
@@ -0,0 +1,47 @@
...
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_201_CREATED)
+
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+
+
+@api_view(["GET", "PUT", "DELETE"])
+def movie_info(request, pk):
```
TPolina, line 28: `movie_detail` is better

TPolina (changes requested):
Delete `.idea` from Github

cinema/views.py
Yaroslav-Stetsko: fixed

TPolina (changes requested):
I don't see fixes)

TPolina (changes requested):
`.idea` is not deleted

TPolina (approved):
Good job!

PR #113, opened 2022-11-16
Solution

Ilyakson:


.gitignore
```diff
@@ -0,0 +1,344 @@
...
+
+# Cython debug symbols
+
+# PyCharm
+#  JetBrains specific template is maintained in a separate JetBrains.gitignore that can
+#  be found at https://github.com/github/gitignore/blob/main/Global/JetBrains.gitignore
+#  and can be added to the global gitignore or merged into this file.  For a more nuclear
+#  option (not recommended) you can uncomment the following to ignore the entire idea folder.
+
+# End of https://www.toptal.com/developers/gitignore/api/pycharm+all,python,django
```
mltdll, line 344: It's better to always end text files with a newline character. You can read why this is the case [here](https://stackoverflow.com/a/729795).

A quick example: If you were to add a new entry to this file via `echo .idea/ >> .gitignore`, it would only work properly if there would be a newline at the end.

cinema/serializers.py
```diff
@@ -0,0 +1,23 @@
+from rest_framework import serializers
+
+from cinema.models import Movie
+
+
+class MovieSerializer(serializers.Serializer):
+    id = serializers.IntegerField(read_only=True)
+    title = serializers.CharField(required=True)
```
mltdll, line 8: I guess it's better to specify `max_length` here.

cinema/views.py
```diff
@@ -0,0 +1,46 @@
...
+    if request.method == "GET":
+        serializer = MovieSerializer(movie)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+
+    if request.method == "PUT":
+        serializer = MovieSerializer(data=request.data)
+
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_201_CREATED)
```
mltdll, line 39: `200` would be a better match here.

mltdll (changes requested):

mltdll (approved):

TPolina (approved):
Good job!

PR #114, opened 2022-11-16
Solution

FranchukS:


pavlo-kapitanets (approved):
Nice!

cinema/models.py
```diff
@@ -0,0 +1,10 @@
+from django.db import models
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField(null=True)
```
TPolina, line 6: for description it's better to write null=True, blank=True

cinema/views.py
```diff
@@ -0,0 +1,45 @@
...
+        movie = Movie.objects.get(pk=pk)
+    except Movie.DoesNotExist:
+        return Response(status=status.HTTP_404_NOT_FOUND)
+
+    if request.method == "GET":
+        serializer = MovieSerializer(movie)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+
+    if request.method == "PUT":
+        serializer = MovieSerializer(data=request.data)
```
TPolina, line 37: you need to pass movie object as well

TPolina (changes requested):

TPolina (approved):
Good job!

PR #115, opened 2022-11-17
checklist initial

DmytroSvirsa:


Abnormaltype (changes requested):
use `get_object_or_404` instead of try/except

Abnormaltype (approved):
Well done!

PR #123, opened 2022-12-06
Solution

davidkrivko:


Abnormaltype (changes requested):
read the checklist before pushing your solution.

Serhii-Leonenko (changes requested):
Remove redundant db :
image

Serhii-Leonenko (changes requested):
Db is still here, you should not only add it to  .gitignore. Delete it from your repo `git rm --cached file_to_remove`

image

IvanRamyk (approved):

PR #141, opened 2023-02-01
CRUD operations

antonychumak:
Implemented an API that provides CRUD operations for the Movie model


ArsenPidhoretskyi (approved):
Well done!

PR #152, opened 2023-03-02
my solution

yanpurdenko:


cinema/models.py
```diff
@@ -0,0 +1,10 @@
+from django.db import models
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.CharField(max_length=255)
+    duration = models.IntegerField()
+
+    def __str__(self):
+        return str(self.title)
```
Bohdan-Salamakha, line 10: title is already str u don't need to use str() for concertation
add typehint

Bohdan-Salamakha (changes requested):
v1

Bohdan-Salamakha (changes requested):
v2

cinema/models.py
yanpurdenko: misunderstood your last comment

ArsenPidhoretskyi (approved):
Well done!

PR #167, opened 2023-03-28
Develop

Zoriana-Dvulit:


Zoriana-Dvulit:
image

cinema/models.py
```diff
@@ -0,0 +1,13 @@
...
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    duration = models.PositiveIntegerField()
+
+    class Meta:
+        verbose_name_plural = "movies"
+
+    def __str__(self):
+        return str(self.title)
```
ArsenPidhoretskyi, line 13: Why you need str() here?

cinema/views.py
```diff
@@ -0,0 +1,44 @@
...
+def movie_list(request):
+    if request.method == "GET":
+        movies = Movie.objects.all()
+        serializer = MovieSerializer(movies, many=True)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+
+    if request.method == "POST":
+        serializer = MovieSerializer(data=request.data)
+        if serializer.is_valid():
+            serializer.save()
```
ArsenPidhoretskyi, line 18: use serializer.is_valid(raise_exception=True) instead of if

cinema/views.py
```diff
@@ -0,0 +1,44 @@
...
+
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+
+
+@api_view(["GET", "PUT", "DELETE"])
+def movie_detail(request, pk):
+    try:
+        movie = Movie.objects.get(pk=pk)
+    except Movie.DoesNotExist:
+        return Response(status=status.HTTP_404_NOT_FOUND)
```
ArsenPidhoretskyi, line 29: Use get_object_or_404

cinema/views.py
```diff
@@ -0,0 +1,44 @@
...
+
+    if request.method == "GET":
+        serializer = MovieSerializer(movie)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+
+    elif request.method == "PUT":
+        serializer = MovieSerializer(movie, data=request.data)
+        if serializer.is_valid():
+            serializer.save()
```
ArsenPidhoretskyi, line 38: The same use serializer.is_valid(raise_exception=True) instead of if

ArsenPidhoretskyi (changes requested):
Remove .idea from github

cinema/models.py
```diff
@@ -8,6 +8,3 @@ class Movie(models.Model):
 
     class Meta:
         verbose_name_plural = "movies"
-
-    def __str__(self):
-        return str(self.title)
```
ArsenPidhoretskyi: I mean another
```
def __str__(self):
        return self.title
```

cinema/views.py
```diff
@@ -0,0 +1,42 @@
...
+        serializer = MovieSerializer(movies, many=True)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+
+    if request.method == "POST":
+        serializer = MovieSerializer(data=request.data)
+        if serializer.is_valid(raise_exception=True):
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_201_CREATED)
+
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```
ArsenPidhoretskyi, line 22: For me not good. Did you read documentation how serializer.is_valid(raise_exception=True) works?
Is really needed if statement here? Maybe we can write this just inline?
Fix in other place

ArsenPidhoretskyi (changes requested):
.idea exists in github. Remove it please! Read about git rm --cached

cinema/views.py
Zoriana-Dvulit: Thanks. I have read about it in https://www.django-rest-framework.org/api-guide/serializers/- [Raising an exception on invalid data](https://www.django-rest-framework.org/api-guide/serializers/#raising-an-exception-on-invalid-data)
The .is_valid() method takes an optional raise_exception flag that will cause it to raise a serializers.ValidationError exception if there are validation errors.

These exceptions are automatically dealt with by the default exception handler that REST framework provides, and will return HTTP 400 Bad Request responses by default.

# Return a 400 response if the data was invalid.

db.sqlite3
Bohdan-Salamakha, line 1: u don't need to commit db.sqlite3, remove it

Zoriana-Dvulit:
I created .gitignore file and removed all redundant files, and also today in QnA seession asked mentor Olena about it. I really don't know why these files appear again in GitHub. Thanks.

Bohdan-Salamakha (approved):

PR #171, opened 2023-04-01
'Solution'

natalie-goriela:


ArsenPidhoretskyi (changes requested):

cinema/views.py
```diff
@@ -0,0 +1,60 @@
...
+
+    if request.method == "POST":
+        serializer = MovieSerializer(data=request.data)
+
+        if serializer.is_valid(raise_exception=True):
+            serializer.save()
+            return Response(
+                serializer.data,
+                status=status.HTTP_201_CREATED
+            )
```
EdAlekseiev, line 25: When you use `serializer.is_valid(raise_exception=True)` you don't need to check returned value, because if data is invalid this method will raise ValidationError

cinema/views.py
```diff
@@ -0,0 +1,60 @@
...
+            status=status.HTTP_200_OK
+        )
+
+    if request.method == "PUT":
+        serializer = MovieSerializer(movie, data=request.data)
+        if serializer.is_valid(raise_exception=True):
+            return Response(
+                serializer.data,
+                status=status.HTTP_200_OK
+            )
```
EdAlekseiev, line 51: When you use serializer.is_valid(raise_exception=True) you don't need to check returned value, because if data is invalid this method will raise ValidationError

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Well done

PR #175, opened 2023-04-03
v1

MaxiTDI:


cinema/views.py
```diff
@@ -0,0 +1,41 @@
...
+        movies = Movie.objects.all()
+        serializer = MovieSerializer(movies, many=True)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+
+    if request.method == "POST":
+        serializer = MovieSerializer(data=request.data)
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_201_CREATED)
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```
ArsenPidhoretskyi, line 22: You can use 
serializer.is_valid(raise_exception=True) instead of if statement
Fix in other places

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):
good job!

PR #185, opened 2023-04-04
Solution according to readme and checklist

NEkropo1:


yura-hudzovskyi (approved):
GJ!

cinema/serializers.py
```diff
@@ -0,0 +1,28 @@
...
+    def create(self, validated_data):
+        return Movie.objects.create(**validated_data)
+
+    def update(self, instance, validated_data):
+        instance.title = validated_data.get("title", instance.title)
+        instance.description = validated_data.get(
+            "description",
+            instance.description
+        )
+        instance.duration = serializers.IntegerField()
```
Y-Havryliv, line 26: seems like you are not updating it

cinema/views.py
```diff
@@ -0,0 +1,43 @@
...
+
+    if request.method == "GET":
+        serializer = MovieSerializer(movie)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+
+    if request.method == "PUT":
+        serializer = MovieSerializer(movie, data=request.data)
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_204_NO_CONTENT)
```
Y-Havryliv, line 38: wrong status

Y-Havryliv (changes requested):

cinema/views.py
NEkropo1: You don't need to reload the page after a successful PUT request. When you send a PUT request to update a resource, the server should update the resource and return a response that indicates the update was successful.
The status status.HTTP_200_OK generally indicates that the request was successful and the server is returning a response with requested information. However, in this specific case, since the request is an update operation and not a retrieval operation, it might be more appropriate to return a status.HTTP_204_NO_CONTENT response to indicate that the server has successfully processed the request, but there is no content to return. This indicates to the client that the update was successful and there is no need to refresh or reload the resource.

cinema/serializers.py
NEkropo1: Can't update what we don't have (

cinema/views.py
Y-Havryliv: if it's the answer from gpt then it's wrong)
in API as a rule, after you PUT or PATCH some object, the changed object is returned, so 204 is not good for sure

Y-Havryliv: Screenshot 2023-04-05 at 11 53 05
Screenshot 2023-04-05 at 11 52 53
You can see what statuses DRF returns on put and patch

Y-Havryliv (approved):
Nice!

cinema/views.py
NEkropo1: Yup I even checked with most API, which were listed to me by GPT and every returned 200 instead of 204 declaimed by him... Well, at least now I know that xD

PR #186, opened 2023-04-04
'Solution'

Vitalii2906:


cinema/views.py
```diff
@@ -0,0 +1,53 @@
...
+    if request.method == 'GET':
+        movies = Movie.objects.all()
+        serializer = MovieSerializer(movies, many=True)
+        return JsonResponse(serializer.data, safe=False)
+
+    elif request.method == 'POST':
+        data = JSONParser().parse(request)
+        serializer = MovieSerializer(data=data)
+        if serializer.is_valid():
+            serializer.save()
```
ArsenPidhoretskyi, line 24: you can remove if-statement, for this use serializer.is_valid(raise_exception=True)
Fix in all places

ArsenPidhoretskyi (changes requested):
Remove .idea from github
Investigate and use git rm --cached

cinema/views.py
```diff
@@ -0,0 +1,50 @@
...
+
+@csrf_exempt
+def movie_detail(
+        request: WSGIRequest,
+        pk: int
+) -> JsonResponse | HttpResponse:
+    try:
+        movie = Movie.objects.get(pk=pk)
+    except Movie.DoesNotExist:
+        return HttpResponse(status=404)
```
ArsenPidhoretskyi, line 35: Use get_object or 404

cinema/views.py
```diff
@@ -0,0 +1,50 @@
...
+from cinema.models import Movie
+from cinema.serializers import MovieSerializer
+
+
+@csrf_exempt
+def movie_list(request: WSGIRequest) -> JsonResponse:
+    if request.method == 'GET':
+        movies = Movie.objects.all()
+        serializer = MovieSerializer(movies, many=True)
+        return JsonResponse(serializer.data, safe=False)
```
ArsenPidhoretskyi, line 17: Better to return HttpResponse

ArsenPidhoretskyi: Fin in all

ArsenPidhoretskyi (changes requested):
Also remove database from git

PR #193, opened 2023-04-04
Sulution

MikhailLyvak:


oigod (approved):
Good Job, but there is small problem in urls.py , github shows that u miss a whitespace pls check it!

cinema/models.py
```diff
@@ -0,0 +1,13 @@
+from django.db import models
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField(null=True, blank=True)
+    duration = models.IntegerField()
+
+    class Meta:
+        ordering = ["title",]
```
Y-Havryliv, line 10:
```suggestion
        ordering = ["title"]
```

cinema/serializers.py
```diff
@@ -0,0 +1,23 @@
+from rest_framework import serializers
+
+from cinema.models import Movie
+
+
+class MovieSerializer(serializers.Serializer):
+    id = serializers.IntegerField(read_only=True)
+    title = serializers.CharField(required=True)
```
Y-Havryliv, line 8: https://www.django-rest-framework.org/api-guide/fields/#required

cinema/urls.py
```diff
@@ -0,0 +1,13 @@
+from django.urls import path
+
+from .views import (
+    cinema_list,
+    ciname_datail
+)
+
+urlpatterns = [
+    path("home/", cinema_list, name="cinema-list"),
```
Y-Havryliv, line 9: it's not the RESTful API, I can change movies on home enpoint

Y-Havryliv, line 9: also what is cinema_list, is it a good name?

cinema/urls.py
```diff
@@ -0,0 +1,13 @@
...
+    path("home/<int:pk>/", ciname_datail, name="cinema-detail"),
```
Y-Havryliv, line 10: same here

cinema/views.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+@api_view(["GET", "PUT", "DELETE"])
+def ciname_datail(request, pk):
+    movie = get_object_or_404(Movie, pk=pk)
+    if request.method == "GET":
+        serializer = MovieSerializer(movie)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+
+    elif request.method == "PUT":
```
Y-Havryliv, line 34: it's better to use `if`, if you have return above

Y-Havryliv (changes requested):

cinema/serializers.py
YuriiKindrat: didn't fix that

cinema/views.py
YuriiKindrat: same

YuriiKindrat (changes requested):

PR #203, opened 2023-04-04
Solution

ZhAlexR:


cinema/urls.py
```diff
@@ -0,0 +1,10 @@
+from django.urls import path
+
+from cinema.views import movies, change_movie
+
+app_name = "cinema"
+
+urlpatterns = [
+    path("cinema/movies/", movies, name="movie-list"),
+    path("cinema/movies/<pk>/", change_movie, name="change-movie"),
```
YuriiKindrat, line 9: why don't u move similar part to an config/urls.py?

YuriiKindrat (changes requested):

PR #208, opened 2023-04-04
Solution

myroslavatrschk:


skmiwa1 (approved):
I like  what I see!

cinema/views.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+@api_view(["GET", "POST"])
+def movies_list(request):
+    if request.method == "GET":
+        movies = Movie.objects.all()
+        serializer = MovieSerializer(movies, many=True)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+
+    elif request.method == "POST":
```
Y-Havryliv, line 17: it's better to use `if`, if you have return above

Y-Havryliv (approved):
That's nice!

PR #212, opened 2023-04-04
Develop

me1nyk:


cinema/serializers.py
```diff
@@ -0,0 +1,20 @@
+from rest_framework import serializers
+
+from cinema.models import Movie
+
+
+class MovieSerializer(serializers.Serializer):
+    id = serializers.IntegerField(read_only=True)
+    title = serializers.CharField(required=True)
+    description = serializers.CharField(required=True)
+    duration = serializers.IntegerField(required=True)
```
YuriiKindrat, line 10: no need to write required here, it is required by default

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #225, opened 2023-04-10
Solution

ykopatko:


cinema/urls.py
```diff
@@ -0,0 +1,17 @@
+from django.urls import path
+from cinema.views import MovieListCreateView, MovieRetrieveUpdateDestroyView
+
+urlpatterns = [
+    path(
+        "cinema/movies/",
```
YuriiKindrat, line 6: so this will be an url like this api/cinema/cinema/movies ?)

YuriiKindrat (changes requested):

cinema/urls.py
ykopatko: It was something wrong with my commits, because I created a new project and commited to this repositery from 2 different projects, because I deleted one of them trying to use git rm. 

It seems, that I have managed to fix that.

PR #230, opened 2023-04-26
Develop

U-Ra-All:


cinema/views.py
```diff
@@ -0,0 +1,72 @@
...
+
+            return Response(
+                serializer.data,
+                status=status.HTTP_201_CREATED,
+            )
+
+        return Response(
+            serializer.errors,
+            status=status.HTTP_400_BAD_REQUEST
+        )
```
Bohdan-Salamakha, line 36: use is_valid(raise_exception=True) to do all checks automatically and raise handled errors if it's needed

cinema/views.py
```diff
@@ -0,0 +1,72 @@
...
+
+            return Response(
+                serializer.data,
+                status=status.HTTP_200_OK,
+            )
+
+        return Response(
+            serializer.errors,
+            status=status.HTTP_400_BAD_REQUEST
+        )
```
Bohdan-Salamakha, line 65: use is_valid(raise_exception=True) to do all checks automatically and raise handled errors if it's needed

Bohdan-Salamakha (changes requested):
v1

cinema/views.py
```diff
@@ -0,0 +1,72 @@
...
+
+            return Response(
+                serializer.data,
+                status=status.HTTP_201_CREATED,
+            )
+
+        return Response(
+            serializer.errors,
+            status=status.HTTP_400_BAD_REQUEST
+        )
```
Bohdan-Salamakha, line 36: u don't need to check if serializer is valid using raise_exception=True
just write it without checking and response 400
https://stackoverflow.com/questions/40348981/django-rest-framework-raise-exception-true

cinema/views.py
```diff
@@ -0,0 +1,72 @@
...
+
+            return Response(
+                serializer.data,
+                status=status.HTTP_200_OK,
+            )
+
+        return Response(
+            serializer.errors,
+            status=status.HTTP_400_BAD_REQUEST
+        )
```
Bohdan-Salamakha, line 65: u don't need to check if serializer is valid using raise_exception=True
just write it without checking and response 400
https://stackoverflow.com/questions/40348981/django-rest-framework-raise-exception-true

Bohdan-Salamakha (changes requested):
v2

cinema/views.py
U-Ra-All: Aha, now I've got it

ArsenPidhoretskyi (approved):
Well done!

U-Ra-All:
> Well done!

Thanks!

PR #231, opened 2023-04-27
Solution

StreetNik:


cinema/views.py
```diff
@@ -0,0 +1,49 @@
...
+        return Response(serializer.data, status=status.HTTP_200_OK)
+
+    if request.method == "POST":
+        serializer = MovieSerializer(data=request.data)
+
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer, status.HTTP_201_CREATED)
+
+        return Response(serializer.errors, status.HTTP_400_BAD_REQUEST)
```
ArsenPidhoretskyi, line 25: Instead of if-else use 

ArsenPidhoretskyi: Fix in all places

cinema/views.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+@api_view(["GET", "PUT", "DELETE"])
+def bus_info(request, pk):
```
ArsenPidhoretskyi, line 29: image
The same error with list
Rename it to movie

cinema/views.py
```diff
@@ -0,0 +1,49 @@
...
+    try:
+        movie = Movie.objects.get(id=pk)
+    except Movie.DoesNotExist:
+        return HttpResponse(status=404)
```
ArsenPidhoretskyi, line 33: Use get_object_or_404 instead of try-except

ArsenPidhoretskyi (changes requested):
Remove .idea from github
Also remove `__pycache__`

ArsenPidhoretskyi (approved):
Well done! Also do not forget about type annotations

PR #232, opened 2023-05-01
Develop

alina-boichenko:


cinema/views.py
```diff
@@ -0,0 +1,50 @@
...
+        serializer = MovieSerializer(movies, many=True)
+
+        return Response(serializer.data, status=status.HTTP_200_OK)
+
+    if request.method == "POST":
+        serializer = MovieSerializer(data=request.data)
+
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_201_CREATED)
```
ArsenPidhoretskyi, line 24: Use serializer.is_valid(raise_exception=True) instead of if statement
Fix in all places

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #237, opened 2023-05-11
done

phaishuk:


cinema/serializers.py
```diff
@@ -0,0 +1,19 @@
...
+    duration = serializers.IntegerField(required=True)
+
+    def create(self, validated_data):
+        return Movie.objects.create(**validated_data)
+
+    def update(self, instance, validated_data):
+        instance.title = validated_data.get("title", instance.title)
+        instance.description = validated_data.get("description", instance.description)
+        instance.duration = validated_data.get("duration", instance.duration)
+        return instance
```
olena-hul, line 19: Missing .save() method for instance

cinema/views.py
```diff
@@ -0,0 +1,44 @@
...
+def movie_list(request):
+    if request.method == "GET":
+        movies = Movie.objects.all()
+        serializer = MovieSerializer(movies, many=True)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+
+    if request.method == "POST":
+        serializer = MovieSerializer(data=request.data)
+
+        if serializer.is_valid():
```
olena-hul, line 20:
```suggestion
        serializer.is_valid(raise_exception=True):
```

This way you won't need to return 400 response by yourself

olena-hul, line 20:
```suggestion
        serializer.is_valid(raise_exception=True):
```

This way you won't need to return 400 response by yourself

olena-hul (changes requested):

cinema/views.py
```diff
@@ -0,0 +1,41 @@
...
+    if request.method == "GET":
+        serializer = MovieSerializer(movie)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+
+    elif request.method == "PUT":
+        serializer = MovieSerializer(movie, data=request.data)
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status.HTTP_200_OK)
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```
ArsenPidhoretskyi, line 37: Better to use serializer.is_valid(raise_exception=True) instead of if-else

ArsenPidhoretskyi (changes requested):

cinema/views.py
phaishuk: forget about this line

ArsenPidhoretskyi (approved):
Well done!

PR #239, opened 2023-05-17
Solution

Oomamchur:


cinema/views.py
```diff
@@ -0,0 +1,41 @@
...
+@api_view(["GET", "POST"])
+def movie_list(request) -> Response:
+    if request.method == "GET":
+        movies = Movie.objects.all()
+        serializer = MovieSerializer(movies, many=True)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+
+    if request.method == "POST":
+        serializer = MovieSerializer(data=request.data)
+        if serializer.is_valid():
```
olena-hul, line 19:
```suggestion
        serializer.is_valid(raise_exception=True):

```
In this way the 400 response will be returned automatically when the data is not valid, and you don't need to return it by yourself

olena-hul: Fix everywhere

olena-hul (changes requested):

cinema/views.py
```diff
@@ -0,0 +1,38 @@
...
+    if request.method == "GET":
+        movies = Movie.objects.all()
+        serializer = MovieSerializer(movies, many=True)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+
+    if request.method == "POST":
+        serializer = MovieSerializer(data=request.data)
+        if serializer.is_valid(raise_exception=True):
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_201_CREATED)
```
Bohdan-Salamakha, line 21: use is_valid(raise_exception=True) to do all checks automatically and raise handled errors if it's needed

cinema/views.py
```diff
@@ -0,0 +1,38 @@
...
+    movie = get_object_or_404(Movie, id=pk)
+    if request.method == "GET":
+        serializer = MovieSerializer(movie)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+
+    if request.method in ["PUT", "PATCH"]:
+        serializer = MovieSerializer(movie, data=request.data)
+        if serializer.is_valid():
+            serializer.save(raise_exception=True)
+            return Response(serializer.data, status=status.HTTP_200_OK)
```
Bohdan-Salamakha, line 34: use is_valid(raise_exception=True) to do all checks automatically and raise handled errors if it's needed

Bohdan-Salamakha (changes requested):
v1

YukihiroSM (approved):
Looks good now.

PR #240, opened 2023-05-21
Solution Maksym Maksimov

MKeSiMu:


cinema/views.py
```diff
@@ -0,0 +1,47 @@
...
+@api_view(["GET", "POST"])
+def movie_list(request):
+    if request.method == "GET":
+        movies = Movie.objects.all()
+        serializer = MovieSerializer(movies, many=True)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+    if request.method == "POST":
+        serializer = MovieSerializer(data=request.data)
+
+        if serializer.is_valid():
```
Vasyl-Smutok, line 21:
```suggestion
        serializer.is_valid(raise_exception=True):
```

In this way the 400 response will be returned automatically when the data is not valid, and you don't need to return it by yourself

Vasyl-Smutok: Fix everywhere

Vasyl-Smutok (changes requested):

ArsenPidhoretskyi (changes requested):

cinema/serializers.py
```diff
@@ -0,0 +1,22 @@
...
+    id = serializers.IntegerField(read_only=True)
+    title = serializers.CharField()
+    description = serializers.CharField()
+    duration = serializers.IntegerField()
+
+    def create(self, validated_data):
+        return Movie.objects.create(**validated_data)
+
+    def update(self, instance, validated_data):
+        instance.title = validated_data.get("title", instance.title)
```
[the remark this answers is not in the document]
Grayder0152: Your approach is not bad but, let's do something like this to avoid getting and changing attribute always

Vasyl-Smutok (approved):
Good Job 👍

PR #245, opened 2023-05-29
Solution

Soobig666:


cinema/views.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+class MovieList(generics.ListCreateAPIView):
+    queryset = Movie.objects.all()
+    serializer_class = MovieSerializer
+
+
+class MovieDetail(generics.RetrieveUpdateDestroyAPIView):
+    queryset = Movie.objects.all()
+    serializer_class = MovieSerializer
```
EdAlekseiev, line 13: You must use function-based views(api_view)

EdAlekseiev (changes requested):

PR #252, opened 2023-06-09
Solution

destr34:


cinema/views.py
```diff
@@ -0,0 +1,44 @@
...
+        movies = Movie.objects.all()
+        serializer = MovieSerializer(movies, many=True)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+    if request.method == "POST":
+        serializer = MovieSerializer(data=request.data)
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_201_CREATED)
+
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```
olena-hul, line 21: Use `serializer.is_valid(raise_exception=True)` without `if`. It will automatically return 400 instead of handling wrong input manually

Bohdan-Salamakha (approved):

PR #253, opened 2023-06-13
Solution

AlexandrZhezheria:


cinema/views.py
```diff
@@ -0,0 +1,46 @@
...
+def movie_list(request):
+    if request.method == "GET":
+        movies = Movie.objects.all()
+        serializer = MovieSerializer(movies, many=True)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+
+    if request.method == "POST":
+        serializer = MovieSerializer(data=request.data)
+
+        if serializer.is_valid():
```
vasylhnatiuk, line 20: Use serializer.is_valid(raise_exception=True) instead of if statement
Fix here and in all of th code

vasylhnatiuk (approved):

PR #260, opened 2023-06-26
Solution

KittNeKit:


cinema/views.py
```diff
@@ -0,0 +1,45 @@
...
+def movie_list(request):
+    if request.method == "GET":
+        movies = Movie.objects.all()
+        serializer = MovieSerializer(movies, many=True)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+
+    if request.method == "POST":
+        serializer = MovieSerializer(data=request.data)
+
+        if serializer.is_valid():
```
EdAlekseiev, line 20: Use `serializer.is_valid(raise_exception=True)` instead of if-statement

EdAlekseiev (changes requested):

cinema/views.py
```diff
@@ -0,0 +1,42 @@
...
+        movies = Movie.objects.all()
+        serializer = MovieSerializer(movies, many=True)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+
+    if request.method == "POST":
+        serializer = MovieSerializer(data=request.data)
+
+        if serializer.is_valid(raise_exception=True):
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_201_CREATED)
```
EdAlekseiev, line 22:
```suggestion
        serializer.is_valid(raise_exception=True):
        serializer.save()
        return Response(serializer.data, status=status.HTTP_201_CREATED)
```

EdAlekseiev (changes requested):

PR #262, opened 2023-06-26
Created all but have an error  No fixture named 'cinema_service_db_da…

vlamuha:
…ta' found.

### DIFF
--- a/cinema/views.py
+++ b/cinema/views.py
@@ -17,11 +17,9 @@ def movie_list(request):
     if request.method == "POST":
         serializer = MovieSerializer(data=request.data)
 
-        if serializer.is_valid():
-            serializer.save()
-            return Response(serializer.data, status=status.HTTP_201_CREATED)
-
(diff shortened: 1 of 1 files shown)


cinema/views.py
```diff
@@ -0,0 +1,44 @@
...
+def movie_list(request):
+    if request.method == "GET":
+        movies = Movie.objects.all()
+        serializer = MovieSerializer(movies, many=True)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+
+    if request.method == "POST":
+        serializer = MovieSerializer(data=request.data)
+
+        if serializer.is_valid():
```
ArsenPidhoretskyi, line 20: Instead of an if statement, useL serializer.is_valid(raise_exception=True)
Fix in all places

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #265, opened 2023-06-27
Created api by CRUD for movies and movies detail

Sebshe:


cinema/views.py
```diff
@@ -0,0 +1,50 @@
...
+
+        return Response(serializer.data, status=status.HTTP_200_OK)
+
+    def post(self, request):
+        serializer = MovieSerializer(data=request.data)
+
+        if serializer.is_valid():
+            serializer.save()
+
+            return Response(serializer.data, status=status.HTTP_201_CREATED)
```
ArsenPidhoretskyi, line 23: No need to use if statement just do:
Fix in all places

ArsenPidhoretskyi (changes requested):

vsmutok (approved):
Well Done 👍

PR #276, opened 2023-07-12
'Solution'

andy77andy:


vasylhnatiuk (changes requested):
you add 5k+ files, add in gitignore and  delete from commit all of redundant files

vasylhnatiuk (changes requested):
code is good. why do you need Amend file? delete pls

Grayder0152 (approved):

PR #277, opened 2023-07-13
Solution

Bloodviel:


cinema/views.py
```diff
@@ -0,0 +1,68 @@
...
+
+        return Response(
+            serializer.data,
+            status=status.HTTP_200_OK
+        )
+
+    if request.method == "POST":
+        serializer = MovieSerializer(data=request.data)
+
+        if serializer.is_valid():
```
vasylhnatiuk, line 23: use serializer.is_valid(raise_exception=True) instead of this here and in all of the code

vasylhnatiuk (changes requested):

vsmutok (approved):
Looks Good 👍

PR #286, opened 2023-07-19
movie-api

muhactive:


cinema/views.py
```diff
@@ -0,0 +1,44 @@
...
+        movies = Movie.objects.all()
+        serializer = MovieSerializer(movies, many=True)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+
+    if request.method == "POST":
+        serializer = MovieSerializer(data=request.data)
+
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_201_CREATED)
```
EdAlekseiev, line 22: Use `serializer.is_valid(raise_exception=True)` instead of `if-statement` block

cinema/views.py
```diff
@@ -0,0 +1,44 @@
...
+
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```
EdAlekseiev, line 24:
```suggestion
        serializer.save(raise_exception=True)
        return Response(serializer.data, status=status.HTTP_201_CREATED)
```

cinema/views.py
```diff
@@ -0,0 +1,44 @@
...
+        serializer = MovieSerializer(movie)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+
+    if request.method == "PUT":
+        serializer = MovieSerializer(movie, data=request.data)
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_202_ACCEPTED)
+
+        return Response(serializer.errors, status=status.HTTP_406_NOT_ACCEPTABLE)
```
EdAlekseiev, line 40: The same comment as above

EdAlekseiev (changes requested):

EdAlekseiev (changes requested):
V1

EdAlekseiev (approved):
Good job

PR #290, opened 2023-07-26
implemented an API that provides CRUD operations for the Movie model

N1khto:


cinema/views.py
```diff
@@ -0,0 +1,44 @@
...
+def movie_list(request):
+    if request.method == "GET":
+        movies = Movie.objects.all()
+        serializer = MovieSerializer(movies, many=True)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+
+    if request.method == "POST":
+        serializer = MovieSerializer(data=request.data)
+
+        if serializer.is_valid():
```
ArsenPidhoretskyi, line 20: better to use serializer.is_valid(raise_exception=True) instead of if-else statement

ArsenPidhoretskyi (changes requested):

cinema/views.py
```diff
@@ -0,0 +1,41 @@
...
+def movie_detail(request, pk):
+    movie = get_object_or_404(Movie, pk=pk)
+
+    if request.method == "GET":
+        serializer = MovieSerializer(movie)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+
+    elif request.method == "PUT":
+        serializer = MovieSerializer(movie, data=request.data)
+        if serializer.is_valid():
```
vasylhnatiuk, line 34: serializer.is_valid(raise_exception=True) use here also

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #294, opened 2023-07-27
Solution #1

erikagayan:


.gitignore
```diff
@@ -0,0 +1,4 @@
+.idea/
+venv/
+**__pycache__/
+db.sqlite3
```
YuriiKindrat, line 4: no blank here (

cinema/models.py
```diff
@@ -0,0 +1,13 @@
+from django.db import models
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255, null=True)
+    description = models.TextField()
+    duration = models.PositiveIntegerField()
+
+    class Meta:
+        verbose_name_plural = "movies"
```
YuriiKindrat, line 10: why do u need this ?
image

cinema/views.py
```diff
@@ -0,0 +1,41 @@
...
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_201_CREATED)
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+
+@api_view(['GET', 'PUT', 'DELETE'])
+def movie_detail(request, pk):
+    try:
+        movie = Movie.objects.get(pk=pk)
+    except Movie.DoesNotExist:
+        return Response(status=status.HTTP_404_NOT_FOUND)
```
YuriiKindrat, line 28: could u use method get_or_404 here?

YuriiKindrat (changes requested):

.gitignore
YuriiKindrat: u didn't fix that

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #297, opened 2023-07-27
Solution

volodymyr-komarnyckyi:


cinema/views.py
```diff
@@ -0,0 +1,45 @@
...
+
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+
+
+@api_view(["GET", "PUT", "DELETE"])
+def movie_detail(request, pk):
+    try:
+        movie = Movie.objects.get(pk=pk)
+    except Movie.DoesNotExist:
+        return Response(status=status.HTTP_400_BAD_REQUEST)
```
YuriiKindrat, line 30: u can use get_or_404 method here instead

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #329, opened 2023-08-14
Create Cinema API

Bohdan11Dii:


cinema/views.py
```diff
@@ -0,0 +1,45 @@
...
+        movies = Movie.objects.all()
+        serializer = MovieSerializers(movies, many=True)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+
+    if request.method == "POST":
+        serializer = MovieSerializers(data=request.data)
+
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_201_CREATED)
```
EdAlekseiev, line 22: Please, use `is_valid` with the argument `raise_exception=True` instead of `if-statement`.

cinema/views.py
```diff
@@ -0,0 +1,45 @@
...
+
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```
EdAlekseiev, line 24:
```suggestion
        serializer.is_valid(raise_exception=True)
        serializer.save()
        return Response(serializer.data, status=status.HTTP_201_CREATED)
```

cinema/views.py
```diff
@@ -0,0 +1,45 @@
...
+        serializer = MovieSerializers(movie)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+
+    elif request.method == "PUT":
+        serializer = MovieSerializers(movie, data=request.data)
+
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_201_CREATED)
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```
EdAlekseiev, line 41: The same comment as above

EdAlekseiev (changes requested):

cinema/views.py
Bohdan11Dii: Why is it more correct to use it than if-statement?

PR #339, opened 2023-09-07
Py-movie-api

alexgubarev88:


cinema/views.py
```diff
@@ -0,0 +1,42 @@
...
+    if request.method == "GET":
+        serializer = MovieSerializer(movie)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+
+    elif request.method == "PUT":
+        serializer = MovieSerializer(movie, data=request.data)
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_201_CREATED)
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```
ArsenPidhoretskyi, line 38: Use serializer.is_valid(raise_exception=True) instead of if statement

ArsenPidhoretskyi (changes requested):

Yurnerroo (changes requested):
if statement is not needed: 
`if serializer.is_valid(raise_exception=True):` 
just call `serializer.is_valid(raise_exception=True)` instead

vasylhnatiuk (approved):

PR #343, opened 2023-09-26
Solved

Dimo-n:


cinema/models.py
```diff
@@ -0,0 +1,13 @@
...
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    duration = models.IntegerField()
+
+    class Meta:
+        verbose_name_plural = "movies"
+
+    def __str__(self):
+        return str(self.title)
```
pavlejviki, line 13: self.title is already a string

cinema/views.py
```diff
@@ -0,0 +1,44 @@
...
+
+    if request.method == "GET":
+        serializer = MovieSerializer(movie)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+
+    if request.method == "PUT":
+        serializer = MovieSerializer(movie, data=request.data)
+        serializer.is_valid(raise_exception=True)
+        serializer.save()
+        return Response(serializer.data, status=status.HTTP_204_NO_CONTENT)
```
pavlejviki, line 40: For a PUT request, where you are updating a resource and returning the updated data, it might be more appropriate to return HTTP_200_OK

pavlejviki (changes requested):

pavlejviki (approved):
Well done!

PR #355, opened 2023-10-09
Solution

nyanplague:


cinema/views.py
```diff
@@ -0,0 +1,45 @@
...
+
+    if request.method == "PUT":
+        movie = get_object_or_404(Movie, id=pk)
+        serializer = MovieSerializer(movie, data=request.data)
+        serializer.is_valid(raise_exception=True)
+        serializer.save()
+        return Response(serializer.data, status=status.HTTP_200_OK)
+
+    if request.method == "DELETE":
+        movie = get_object_or_404(Movie, id=pk)
```
lgSilay, line 40: You can use "movie = get_object_or_404(Movie, id=pk)" directly in function and not repeat yourself three times)

lgSilay (approved):
Good job!)

cinema/views.py
nyanplague: @lgSilay thank you :)

PR #372, opened 2023-10-10
soluiton

OleksandrMoroz24:


cinema_project/cinema/serializers.py
```diff
@@ -0,0 +1,8 @@
+from rest_framework import serializers
+from .models import Movie
+
+
+class MovieSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Movie
+        fields = ['id', 'title', 'description', 'duration']
```
Sofbog, line 8: if u r using all fields mb `fields = '__all__'` will be better

Sofbog (changes requested):

Sofbog (approved):
LGTM

PR #376, opened 2023-10-11
Solution

Kurta97:


anna-lybid (approved):

cinema/models.py
```diff
@@ -0,0 +1,13 @@
+from django.db import models
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField(null=True, blank=True)
+    duration = models.PositiveIntegerField(null=True, blank=True)
```
Sofbog, line 7: is it possible to have a movie without duration?

Sofbog (changes requested):

cinema/models.py
Kurta97: The film cannot but have a duration, but we may not know it. 
Or the film has not yet been released. It-s a normal situation if you have site like IMDB

Sofbog (approved):
LGTM

PR #378, opened 2023-10-13
Done

maxkatkalov:


cinema/views.py
```diff
@@ -0,0 +1,47 @@
...
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_201_CREATED)
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+
+
+@api_view(["GET", "PUT", "DELETE"])
+def movie_detail(request, pk: int):
+    try:
+        movie_single_object = Movie.objects.get(pk=pk)
```
Sofbog, line 28: Make sure you catch an error in @api_view in case that object doesn't exist. Use get_object_or_404 instead of try/except for this purpose.

Sofbog (changes requested):

Sofbog (approved):
LGTM

PR #391, opened 2023-11-14
Solution

olmatiash:


cinema/views.py
```diff
@@ -0,0 +1,41 @@
...
+@api_view(["GET", "POST"])
+def movie_list(request):
+    if request.method == "GET":
+        movies = Movie.objects.all()
+        serializer = MovieSerializer(movies, many=True)
+        return Response(serializer.data, status=200)
+
+    if request.method == "POST":
+        serializer = MovieSerializer(data=request.data)
+        if serializer.is_valid():
```
ArsenPidhoretskyi, line 18: Instead of usage if statement use

cinema/views.py
```diff
@@ -0,0 +1,41 @@
...
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_201_CREATED)
+    return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+
+
+@api_view(["GET", "PUT", "DELETE"])
+def movie_detail(request, pk):
+    movie = Movie.objects.get(id=pk)
+    serializer = MovieSerializer(movie)
```
ArsenPidhoretskyi, line 27: Move this into get method

cinema/views.py
ArsenPidhoretskyi: Fix in all places

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):

PR #393, opened 2023-11-18
implemented py-movie-api task

HlovaOleh:


cinema/models.py
```diff
@@ -0,0 +1,10 @@
+from django.db import models
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    duration = models.IntegerField()
+
+    def __str__(self) -> str:
+        return str(self.title)
```
ArsenPidhoretskyi, line 10: Why do you need str()

cinema/views.py
```diff
@@ -0,0 +1,47 @@
...
+def movie_list(request) -> Response:
+    if request.method == "GET":
+        movies = Movie.objects.all()
+        serializer = MovieSerializer(movies, many=True)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+
+    if request.method == "POST":
+        serializer = MovieSerializer(data=request.data)
+
+        if serializer.is_valid():
```
ArsenPidhoretskyi, line 19: Instead of is statement use: serializer.is_valid(raise_exception=True)

ArsenPidhoretskyi (changes requested):

cinema/views.py
```diff
@@ -0,0 +1,47 @@
...
+        return Response(serializer.data, status=status.HTTP_200_OK)
+
+    if request.method == "POST":
+        serializer = MovieSerializer(data=request.data)
+
+        if serializer.is_valid(raise_exception=True):
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_201_CREATED)
+
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```
NikOneZ1, line 23: You don't need this if statement, because is_valid with raise exception will automatically return status 400 if serialiser is not valid
```suggestion
        serializer.is_valid(raise_exception=True):
        serializer.save()
        return Response(serializer.data, status=status.HTTP_201_CREATED)
```

cinema/views.py
```diff
@@ -0,0 +1,47 @@
...
+        return Response(serializer.data, status=status.HTTP_200_OK)
+
+    if request.method == "PUT":
+        serializer = MovieSerializer(movie, data=request.data)
+
+        if serializer.is_valid(raise_exception=True):
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_200_OK)
+
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```
NikOneZ1, line 41: You don't need this if statement, because is_valid with raise exception will automatically return status 400 if serialiser is not valid
```suggestion
        serializer.is_valid(raise_exception=True):
        serializer.save()
        return Response(serializer.data, status=status.HTTP_200_OK)
```

NikOneZ1 (changes requested):

EdAlekseiev (approved):
Good job

PR #399, opened 2023-11-25
Solution/yaroslav-lysiuk

Quiet-Klirik:


cinema/views.py
```diff
@@ -0,0 +1,65 @@
...
+            if self.model.objects.filter(title=new_title).exists():
+                return Response(
+                    {
+                        "error": (f"Object with title '{new_title}' "
+                                  f"is already exists")
+                    },
+                    status=status.HTTP_400_BAD_REQUEST
+                )
+        serializer = self.serializer_class(movie, data=request.data)
+        if serializer.is_valid():
```
ArsenPidhoretskyi, line 54:
```suggestion
        serializer.is_valid(raise_exception=True)

```
instead of if statement use raise_exception=True

cinema/views.py
```diff
@@ -0,0 +1,65 @@
...
+        new_title = request.data.get("title")
+        if not new_title == movie.title:
+            if self.model.objects.filter(title=new_title).exists():
+                return Response(
+                    {
+                        "error": (f"Object with title '{new_title}' "
+                                  f"is already exists")
+                    },
+                    status=status.HTTP_400_BAD_REQUEST
+                )
```
ArsenPidhoretskyi, line 52: Validation is redundant, just change into model unique=True

cinema/views.py
```diff
@@ -0,0 +1,65 @@
...
+            serializer = self.serializer_class(movies, many=True)
+        return Response(serializer.data, status.HTTP_200_OK)
+
+    def post(self, request, *args, **kwargs):
+        title = request.data.get("title")
+        if self.model.objects.filter(title=title).exists():
+            return Response(
+                {"error": f"Object with title '{title}' is already exists"},
+                status=status.HTTP_400_BAD_REQUEST
+            )
```
ArsenPidhoretskyi, line 34: Validation is redundant, just change into model unique=True

ArsenPidhoretskyi (changes requested):

cinema/views.py
Quiet-Klirik: Serializers does not handle UNIQUE errors by default

cinema/views.py
Quiet-Klirik: Serializers does not handle UNIQUE errors by default

cinema/views.py
```diff
@@ -0,0 +1,47 @@
...
+from cinema.serializers import MovieSerializer
+
+
+class MovieCRUDView(APIView):
+    model = Movie
+    serializer_class = MovieSerializer
+
+    def get_object(self, pk: int):
+        instance = get_object_or_404(self.model, pk=pk)
+        return instance
```
Dimosphen1, line 16: There is no need to allocate `instance` variable here, return the output directly from `get_object_or_404`.

cinema/views.py
```diff
@@ -0,0 +1,47 @@
...
+            movie = self.get_object(pk)
+            serializer = self.serializer_class(movie)
+        else:
+            movies = self.model.objects.all()
+            serializer = self.serializer_class(movies, many=True)
+        return Response(serializer.data, status.HTTP_200_OK)
+
+    def post(self, request, *args, **kwargs):
+        serializer = self.serializer_class(data=request.data)
+        if serializer.is_valid(raise_exception=True):
```
Dimosphen1, line 30: There is no need for the `if` condition, remove it and leave the `is_valid` call with `raise_exception=True` parameter. The code will not execute further in case an exception is raised. Fix in all places.

Dimosphen1 (changes requested):
Several changes were requested.

cinema/serializers.py
```diff
@@ -0,0 +1,31 @@
...
+
+class MovieSerializer(serializers.Serializer):
+    id = serializers.IntegerField(read_only=True)
+    title = serializers.CharField(max_length=255)
+    description = serializers.CharField(required=False, allow_blank=True)
+    duration = serializers.IntegerField()
+
+    def validate_title(self, value):
+        old_title = None
+        if self.instance is not None:
```
viktoria-rybenchuk, line 14: Using `if self.instance` is a more concise  way of checking for non-None values.

cinema/views.py
viktoria-rybenchuk: not fixed

cinema/views.py
viktoria-rybenchuk: not fixed

viktoria-rybenchuk (changes requested):

viktoria-rybenchuk (approved):
GJ🚀

PR #416, opened 2023-11-28
Solution

D1st3f:


cinema/models.py
```diff
@@ -0,0 +1,7 @@
+from django.db import models
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField
+    duration = models.IntegerField
```
sakhaline, line 7: Are you sure that the fields are created in your table? Did you remember to call these descriptors ;)

cinema/views.py
```diff
@@ -0,0 +1,43 @@
...
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_201_CREATED)
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+
+
+@api_view(["GET", "PUT", "DELETE"])
+def movie_detail(request, pk):
+    if request.method == "GET":
+        current_movies = Movie.objects.get(pk)
```
sakhaline, line 28: What if the records for the specified pk won't exist in your table? Maybe it's worth checking for the presence of a record? 
And why do you use "movie" in plural in this view?)

cinema/views.py
```diff
@@ -0,0 +1,43 @@
...
+        serializer = MovieSerializer(current_movies)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+
+    elif request.method == "PUT":
```
sakhaline, line 32: You can use " if " here and below too.

sakhaline (changes requested):

sakhaline (approved):
Good job, mate! 🍉

cinema/models.py
```diff
@@ -0,0 +1,7 @@
+from django.db import models
+
+
+class Movie(models.Model):
```
Y-Havryliv, line 4: it's good practice to add the `str` method to models

cinema/serializers.py
```diff
@@ -0,0 +1,20 @@
+from rest_framework import serializers
+
+from cinema.models import Movie
+
+
+class MovieSerializer(serializers.Serializer):
+    id = serializers.IntegerField(read_only=True)
+    title = serializers.CharField(max_length=255)
+    description = serializers.CharField(max_length=3000)
```
Y-Havryliv, line 9: you didn't specify it in a model

cinema/views.py
```diff
@@ -0,0 +1,44 @@
...
+@api_view(["GET", "POST"])
+def movie_list(request) -> Response:
+    if request.method == "GET":
+        movies = Movie.objects.all()
+        serializer = MovieSerializer(movies, many=True)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+
+    elif request.method == "POST":
+        serializer = MovieSerializer(data=request.data)
+        print(serializer.is_valid())
```
Y-Havryliv, line 18: I guess you don't need this print

cinema/views.py
```diff
@@ -0,0 +1,44 @@
...
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_201_CREATED)
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+
+
+@api_view(["GET", "PUT", "DELETE"])
+def movie_detail(request, pk):
+    try:
+        current_movie = Movie.objects.get(pk=pk)
```
Y-Havryliv, line 28: you can use `get_object_or_404`

cinema/views.py
```diff
@@ -0,0 +1,44 @@
...
+        return Response(status=status.HTTP_404_NOT_FOUND)
+    if request.method == "GET":
+        serializer = MovieSerializer(current_movie)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+
+    elif request.method == "PUT":
+        serializer = MovieSerializer(current_movie,data=request.data)
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_201_CREATED)
```
Y-Havryliv, line 39: PUT does not create movies

Y-Havryliv (changes requested):

cinema/models.py
```diff
@@ -0,0 +1,10 @@
...
+    title = models.CharField(max_length=255)
+    description = models.TextField(max_length=3000)
```
Y-Havryliv, line 6: TextField is used for large arbitrary-length strings. You may specify a max_length for the field, but this is used only when the field is displayed in forms (it is not enforced at the database level)
Just in case)

Y-Havryliv (approved):
GJ!

PR #459, opened 2024-02-01
solution

antonnech2309:


cinema/views.py
```diff
@@ -0,0 +1,41 @@
...
+    if request.method == "GET":
+        movies = Movie.objects.all()
+        serializer = MovieSerializer(movies, many=True)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+
+    elif request.method == "POST":
+        serializer = MovieSerializer(data=request.data)
+        if serializer.is_valid(raise_exception=True):
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_201_CREATED)
```
Dimosphen1, line 22: There is no need to place `serializer.is_valid(raise_exception=True)` in the if condition, because it will not allow further code execution in case of error anyway. Replace cases of using this method with `raise_exception` with the following approach 
```suggestion
        serializer = MovieSerializer(data=request.data)
        serializer.is_valid(raise_exception=True)
        serializer.save()
        return Response(serializer.data, status=status.HTTP_201_CREATED)
```

Dimosphen1 (changes requested):
Several minor changes were requested.

PR #487, opened 2024-03-05
Develop

DaniilHinzburh:


vasylhnatiuk (changes requested):
delete .idea from repo
image

cinema/views.py
```diff
@@ -0,0 +1,39 @@
...
+
+@api_view(['GET', 'POST'])
+def movie_list(request):
+    if request.method == 'GET':
+        movies = Movie.objects.all()
+        serializer = MovieSerializer(movies, many=True)
+        return Response(serializer.data, status=200)
+    elif request.method == 'POST':
+        serializer = MovieSerializer(data=request.data)
+        if serializer.is_valid(raise_exception=True):
```
vasylhnatiuk, line 18:
```suggestion
        serializer.is_valid(raise_exception=True)
```

cinema/views.py
```diff
@@ -0,0 +1,39 @@
...
+            serializer.save()
+            return Response(data=serializer.data, status=201)
+        else:
+            return Response(serializer.errors, status=400)
```
vasylhnatiuk, line 22: you don't need it because  `serializer.is_valid(raise_exception=True)` automatically responces with status 400 if serializer is not valid

vasylhnatiuk (changes requested):

cinema/views.py
vasylhnatiuk: do it

cinema/views.py
vasylhnatiuk: do it

vasylhnatiuk (changes requested):
image
delete dbsqlite3 as well and resolve previous comments

vsmutok (approved):
Looks Good!

PR #502, opened 2024-03-13
Task Solution

shaiduchyk:


AntonSkyrda (approved):
use Render from rest framefork instead JsonRender

PR #506, opened 2024-03-13
Develop

artemgrishko:


cinema/urls.py
```diff
@@ -0,0 +1,10 @@
+from django.urls import path
+
+from cinema.views import movie_list, movie_detail
+
+app_name = "cinema"
+
+urlpatterns = [
+    path("movies/", movie_list, name="movie_list"),
+    path("movies/<int:pk>/", movie_detail, name="movie_detail")
+]
```
msymonovych, line 10: I think you should replace "_" with "-" in name

msymonovych (approved):

PR #511, opened 2024-03-13
Develop

MaxOperchuk:


Serejka21 (approved):
All is good, but delete redundant __pycache__ files using git rm

PR #512, opened 2024-03-13
Solution

Serejka21:


movieapi/cinema/views.py
```diff
@@ -0,0 +1,36 @@
+from django.shortcuts import get_object_or_404
+from rest_framework.decorators import api_view
+from rest_framework.response import Response
+
+from .serializers import MovieSerializer
+from .models import Movie
+
+
+@api_view(["GET", "POST"])
+def movie_list(request):
```
mdubyna, line 10: mb put return annotation for functions

movieapi/cinema/views.py
```diff
@@ -0,0 +1,36 @@
...
+def movie_detail(request, pk: int):
+    movie = get_object_or_404(Movie, pk=pk)
+    if request.method == "GET":
+        serializer = MovieSerializer(movie)
+        return Response(serializer.data, status=200)
+    elif request.method == "PUT":
+        serializer = MovieSerializer(movie, data=request.data)
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=200)
```
mdubyna, line 33: if data isn't valid u can return response with errors and status

mdubyna (changes requested):
Add response to PUT method in view, if data isn't valid

PR #524, opened 2024-03-27
Develop

MAFIOZNIK1999:


.idea/.gitignore
LiudmylaKulinchenko, line 1: remove unnecessary files from pr

api/settings.py
```diff
@@ -0,0 +1,125 @@
...
+# Application definition
+
+INSTALLED_APPS = [
+    "django.contrib.admin",
+    "django.contrib.auth",
+    "django.contrib.contenttypes",
+    "django.contrib.sessions",
+    "django.contrib.messages",
+    "django.contrib.staticfiles",
+    "cinema",
```
LiudmylaKulinchenko, line 40: better to move your custom apps to the last place here

cinema/views.py
```diff
@@ -0,0 +1,47 @@
...
+from cinema.serializers import MovieSerializer
+
+
+@api_view(["GET", "POST"])
+def movie_list(request):
+    if request.method == "GET":
+        movies = Movie.objects.all()
+        serializer = MovieSerializer(movies, many=True)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+    else:
```
LiudmylaKulinchenko, line 16: Use `if request.method == "POST":` here

cinema/views.py
```diff
@@ -0,0 +1,47 @@
...
+    elif request.method == "PUT":
+        serializer = MovieSerializer(movie, data=request.data)
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_200_OK)
+        return Response(
+            serializer.errors,
+            status=status.HTTP_400_BAD_REQUEST
+        )
+    else:
```
LiudmylaKulinchenko, line 45: Same here for DELETE

LiudmylaKulinchenko (changes requested):
Remove venv and other unnecessary files from the pr

api/settings.py
LiudmylaKulinchenko: missed my comment 👀

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Well done! 🔥

PR #539, opened 2024-04-20
No message

stamaksim:


cinema/views.py
```diff
@@ -0,0 +1,37 @@
...
+@api_view(["GET", "POST"])
+def movie_list(request) -> Response:
+    if request.method == "GET":
+        movies = Movie.objects.all()
+        serializer = MovieSerializer(movies, many=True)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+    else:
+        serializer = MovieSerializer(data=request.data)
+        if serializer.is_valid():
+            serializer.save()
```
vsmutok, line 18: you can use raise_exception=True in order to avoid return Bad Request. Read [here](https://stackoverflow.com/questions/40348981/django-rest-framework-raise-exception-true)

vsmutok (changes requested):

cinema/views.py
```diff
@@ -14,10 +14,9 @@ def movie_list(request) -> Response:
         return Response(serializer.data, status=status.HTTP_200_OK)
     else:
         serializer = MovieSerializer(data=request.data)
-        if serializer.is_valid():
+        if serializer.is_valid(raise_exception=True):
```
vasylhnatiuk, line 17: you don't need `if` if you use raise_exception=True

vasylhnatiuk (changes requested):

.gitignore
```diff
@@ -0,0 +1,4 @@
+.sqlite3
+.db3
+.pyc
+.idea/
```
vasylhnatiuk, line 4: add newline

vasylhnatiuk (changes requested):

PR #545, opened 2024-04-23
Solution

Bujhm99:


api/urls.py
```diff
@@ -0,0 +1,7 @@
+from django.contrib import admin
+from django.urls import path, include
+
```
MNDonut, line 3: 2 blank lines after the last import

cinema/models.py
```diff
@@ -0,0 +1,10 @@
+from django.db import models
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.CharField(max_length=255, null=True)
```
MNDonut, line 6: it's better to use TextField because you cannot be sure that description will fit in 255 characters, then this field can be nullable and no need to use required=False in serializer

cinema/views.py
```diff
@@ -0,0 +1,42 @@
...
+from cinema.serializers import MovieSerializer
+
+
+@api_view(["GET", "POST"])
+def movies_list(request):
+    if request.method == "GET":
+        movies = Movie.objects.all()
+        serializer = MovieSerializer(movies, many=True)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+    else:
```
MNDonut, line 17: you can remove "else" because you view supports GET and POST only, thus if it isn't get, then you know it's post

cinema/views.py
```diff
@@ -0,0 +1,42 @@
...
+        serializer = MovieSerializer(data=request.data)
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_201_CREATED)
```
MNDonut, line 21: 1 blank line

cinema/views.py
```diff
@@ -0,0 +1,42 @@
...
+        serializer = MovieSerializer(movie)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+    elif request.method == "PUT":
+        serializer = MovieSerializer(movie, data=request.data)
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_200_OK)
+        return Response(
+            serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+    else:
```
MNDonut, line 39: the same, you can remove "else"

MNDonut (changes requested):

LiudmylaKulinchenko (approved):
Good job 🌸

PR #548, opened 2024-04-28
Develop

MykolaMazh:


cinema/views.py
```diff
@@ -0,0 +1,43 @@
...
+from rest_framework.response import Response
+from rest_framework.decorators import api_view
+from rest_framework import status
+
+
+from cinema.models import Movie
+from cinema.serializers import MovieSerializer
+
+
+def create_movie_or_raise(data, instance=None):
```
Dimosphen1, line 12: The name of the method does not correlate with its functionality, rename it to `serialize_movie` or other

cinema/views.py
```diff
@@ -0,0 +1,43 @@
...
+    return Response(serializer.data, status=status.HTTP_201_CREATED)
+
+
+@api_view(["GET", "POST"])
+def movie_list(request):
+    if request.method == "GET":
+        movies = Movie.objects.all()
+        serializer = MovieSerializer(movies, many=True)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+    else:
```
Dimosphen1, line 28: There is no need for `else` since the code will not execute after the `return` statement anyway. Please, remove the redundant `else`

cinema/views.py
```diff
@@ -0,0 +1,43 @@
...
+@api_view(["GET", "PUT", "DELETE"])
+def movie_detail(request, pk):
+    movie = get_object_or_404(Movie, pk=pk)
+    if request.method == "GET":
+        serializer = MovieSerializer(movie)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+    if request.method == "PUT":
+        return create_movie_or_raise(request.data, movie)
+    elif request.method == "DELETE":
+        movie = get_object_or_404(Movie, pk=pk)
```
Dimosphen1, line 41: There is no need to retrieve `movie` here since you already done it on line 34

Dimosphen1 (changes requested):
Several changes were requested

cinema/views.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+@api_view(["GET", "PUT", "DELETE"])
+def movie_detail(request, pk):
+    movie = get_object_or_404(Movie, pk=pk)
+    if request.method == "GET":
+        serializer = MovieSerializer(movie)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+    if request.method == "PUT":
+        return movie_create_or_update(request.data, movie)
```
Dimosphen1, line 38: It would be better to separate `Movie` creation and and update between `POST` and `PUT` methods correspondingly and to not delegate both actions to `PUT`.  Follow this approach in the future

Dimosphen1 (approved):
LGTM, one comment is left for consideration

PR #555, opened 2024-05-09
Solution

twonumberfortyfives:


cinema/views.py
```diff
@@ -0,0 +1,38 @@
...
+@api_view(["GET", "POST"])
+def movie_list(request):
+    if request.method == "GET":
+        movies = Movie.objects.all()
+        serializer = CinemaSerializer(movies, many=True)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+    elif request.method == "POST":
+        serializer = CinemaSerializer(data=request.data)
+        if serializer.is_valid():
+            serializer.save()
```
vsmutok, line 19: Use serializer.is_valid(raise_exception=True) to simplify code. When you use raise_exception=True, if the serializer is invalid, an exception (serializers.ValidationError) is raised. DRF catches this exception and automatically returns a 400 Bad Request response with the appropriate error messages.

vsmutok (changes requested):
Also Please remove `.idea` from PR

db.sqlite3
Dimosphen1, line 1: Remove DB files from the PR

cinema/views.py
```diff
@@ -0,0 +1,36 @@
...
+def movie_detail(request, pk):
+    movie = get_object_or_404(Movie, pk=pk)
+    if request.method == "GET":
+        serializer = CinemaSerializer(movie)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+    elif request.method == "PUT":
+        serializer = CinemaSerializer(movie, data=request.data)
+        if serializer.is_valid(raise_exception=True):
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_200_OK)
```
Dimosphen1, line 33: There is no need to create an if condition here, remove it. Apply this change in other places as well
```suggestion
        serializer.is_valid(raise_exception=True)
        serializer.save()
        return Response(serializer.data, status=status.HTTP_200_OK)
```

Dimosphen1 (changes requested):
Several changes were requested

Dimosphen1 (approved):
LGTM, well done!

PR #557, opened 2024-05-12
Solution

vsupruniuk:


cinema/views.py
```diff
@@ -0,0 +1,49 @@
...
+        if serializer.is_valid():
+            serializer.save()
+
+            return Response(serializer.data, status=status.HTTP_201_CREATED)
+
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+
+
+@api_view(["GET", "PUT", "DELETE"])
+def movie_detail(request: Request, pk):
```
LiudmylaKulinchenko, line 30: Add return type annotations and annotation for pk parameter

LiudmylaKulinchenko (changes requested):
Fix quotes & remove django default comments and docstrings everywhere

LiudmylaKulinchenko (approved):
Nice work 🔥🔥

PR #562, opened 2024-05-19
Solution

donbas777:


Dimosphen1 (changes requested):
Do not commit DB file and venv

db.sqlite3
vsmutok, line 1: Please remove DB from PR

vsmutok (changes requested):
.

cinema/views.py
```diff
@@ -0,0 +1,37 @@
...
+@api_view(["GET", "PUT", "DELETE"])
+def movie_detail(request, pk):
+    movie = get_object_or_404(Movie, pk=pk)
+    if request.method == "GET":
+        serializer = MovieSerializer(movie)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+    if request.method == "PUT":
+        serializer = MovieSerializer(movie, data=request.data)
+        if serializer.is_valid():
+            serializer.save()
```
vsmutok, line 32: Remove the if condition and use the `raise_exception=True` flag for the serializer. Follow this approach in other places as well

vsmutok (changes requested):

cinema/views.py
```diff
@@ -0,0 +1,35 @@
...
+    movie = get_object_or_404(Movie, pk=pk)
+    if request.method == "GET":
+        serializer = MovieSerializer(movie)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+    if request.method == "PUT":
+        serializer = MovieSerializer(movie, data=request.data)
+        serializer.is_valid(raise_exception=True)
+        serializer.save()
+        return Response(serializer.data, status=status.HTTP_200_OK)
+    else:
```
Dimosphen1, line 33: It is better to explicitly check for the delete method rather than default to deletion. Follow this approach in the future

Dimosphen1 (approved):
LGTM, well done! One comment is left for consideration

PR #563, opened 2024-05-19
Solution 1, tasks from Readme.md - DONE))

marinaua13:


cinema/views.py
```diff
@@ -0,0 +1,38 @@
...
+@api_view(["GET", "POST"])
+def movie_list(request):
+    if request.method == 'GET':
+        movies = Movie.objects.all()
+        serializer = MovieSerializer(movies, many=True)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+    else:
+        serializer = MovieSerializer(data=request.data)
+        if serializer.is_valid():
+            serializer.save()
```
vsmutok, line 19: use [raise_exception=True ](https://stackoverflow.com/questions/40348981/django-rest-framework-raise-exception-true)parameter to avoid returning Bad Request

cinema/views.py
```diff
@@ -0,0 +1,38 @@
...
+@api_view(["GET", "PUT", "DELETE"])
+def movie_detail(request, pk):
+    movie = get_object_or_404(Movie, pk=pk)
+    if request.method == "GET":
+        serializer = MovieSerializer(movie)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+    elif request.method == "PUT":
+        serializer = MovieSerializer(movie, data=request.data)
+        if serializer.is_valid():
+            serializer.save()
```
vsmutok, line 33: use [raise_exception=True ](https://stackoverflow.com/questions/40348981/django-rest-framework-raise-exception-true)parameter to avoid returning Bad Request

vsmutok (changes requested):

cinema/views.py
```diff
@@ -15,7 +15,7 @@ def movie_list(request):
         return Response(serializer.data, status=status.HTTP_200_OK)
     else:
         serializer = MovieSerializer(data=request.data)
-        if serializer.is_valid():
+        if serializer.is_valid(raise_exception=True):
```
vasylhnatiuk, line 18: you don't need `if` and  return `Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)`. `raise_exception=True)` makes it automatically

vasylhnatiuk (changes requested):

PR #573, opened 2024-05-22
Develop

panicua:


cinema/models.py
```diff
@@ -0,0 +1,10 @@
+from django.db import models
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=127)
+    description = models.TextField(blank=True)
```
LiudmylaKulinchenko, line 6: description shouldn't be nullable. Imagine trying to pick a movie and description is none - sounds bad 😸

cinema/serializers.py
```diff
@@ -0,0 +1,22 @@
+from rest_framework import serializers
+
+from cinema.models import Movie
+
+
+class MovieSerializer(serializers.Serializer):
+    id = serializers.IntegerField(read_only=True)
+    title = serializers.CharField(required=True, max_length=127)
+    description = serializers.CharField(required=False, allow_blank=True)
+    duration = serializers.IntegerField(required=True)
```
LiudmylaKulinchenko, line 10: it's requred=True by default

movie_api/settings.py
```diff
@@ -0,0 +1,127 @@
...
+
+INSTALLED_APPS = [
+    "django.contrib.admin",
+    "django.contrib.auth",
+    "django.contrib.contenttypes",
+    "django.contrib.sessions",
+    "django.contrib.messages",
+    "django.contrib.staticfiles",
+    "cinema",
+    "rest_framework",
```
LiudmylaKulinchenko, line 43: Your custom apps should be last in the list

LiudmylaKulinchenko (changes requested):
Remove django default commends and docstrings, please

movie_api/settings.py
panicua: got it, I thought about it also, but I was checking the django rest framework documentation and it made me think it should be last because of how it's shown here without ```...``` after "rest_framework":
chrome_SJIZ5vq3mL

LiudmylaKulinchenko: I got you 🙌

movie_api/settings.py
```diff
@@ -1,46 +1,27 @@
...
 INSTALLED_APPS = [
     "django.contrib.admin",
     "django.contrib.auth",
     "django.contrib.contenttypes",
     "django.contrib.sessions",
     "django.contrib.messages",
     "django.contrib.staticfiles",
-    "cinema",
     "rest_framework",
+    # project apps
```
LiudmylaKulinchenko, line 23: You can just use blank lines after default, installed and custom apps. Like with imports, no need to add comments

LiudmylaKulinchenko (approved):
Added one comment, but good job overall! 🌸🌷

PR #576, opened 2024-05-23
Solution

oleksandra-shershen:


app/settings.py
```diff
@@ -0,0 +1,125 @@
...
+
+INSTALLED_APPS = [
+    "django.contrib.admin",
+    "django.contrib.auth",
+    "django.contrib.contenttypes",
+    "django.contrib.sessions",
+    "django.contrib.messages",
+    "django.contrib.staticfiles",
+    "cinema",
+    "rest_framework",
```
LiudmylaKulinchenko, line 41: Your custom apps should be last in this list

cinema/models.py
```diff
@@ -0,0 +1,10 @@
+from django.db import models
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    duration = models.IntegerField()
+
+    def __str__(self):
```
LiudmylaKulinchenko, line 9: add return type annotation

LiudmylaKulinchenko (changes requested):
Remove django default comments and docstrings, please

app/settings.py
oleksandra-shershen: Resolved

oleksandra-shershen: Resolved

cinema/models.py
oleksandra-shershen: Resolved

MNDonut (approved):

PR #578, opened 2024-05-23
Solution

alonasorochynska:


cinema/views.py
```diff
@@ -0,0 +1,38 @@
...
+from cinema.serializers import MovieSerializer
+
+
+@api_view(["GET", "POST"])
+def movie_list(request):
+    if request.method == "GET":
+        movies = Movie.objects.all()
+        serializer = MovieSerializer(movies, many=True)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+    else:
```
LiudmylaKulinchenko, line 16: no need to use else here, since you specified available methods in decorator

cinema/views.py
```diff
@@ -0,0 +1,38 @@
...
+    if request.method == "GET":
+        serializer = MovieSerializer(movie)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+    elif request.method == "PUT":
+        serializer = MovieSerializer(movie, data=request.data)
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_200_OK)
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+    else:
```
LiudmylaKulinchenko, line 36: same here

movie_api/settings.py
```diff
@@ -0,0 +1,125 @@
...
+
+INSTALLED_APPS = [
+    "django.contrib.admin",
+    "django.contrib.auth",
+    "django.contrib.contenttypes",
+    "django.contrib.sessions",
+    "django.contrib.messages",
+    "django.contrib.staticfiles",
+    "cinema",
+    "rest_framework",
```
LiudmylaKulinchenko, line 41: Your custom apps should be last in this list

LiudmylaKulinchenko (changes requested):

MNDonut (approved):

PR #580, opened 2024-05-23
Solution 1

semermaks:


.idea/.gitignore
LiudmylaKulinchenko, line 1: remove idea from github

cinema/midels.py
```diff
@@ -0,0 +1,10 @@
+from django.db import models
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=100)
+    description = models.CharField(null=True, blank=True, max_length=500)
```
LiudmylaKulinchenko, line 6: this fields shouldn't be nullable

cinema/midels.py
```diff
@@ -0,0 +1,10 @@
...
+    duration = models.IntegerField()
+
+    class Meta:
+        app_label = "cinema"
```
LiudmylaKulinchenko, line 10: add str method

cinema/serializers.py
```diff
@@ -0,0 +1,21 @@
+from rest_framework import serializers
+from cinema.models import Movie
+
+
+class MovieSerializer(serializers.Serializer):
+    id = serializers.IntegerField(read_only=True)
+    title = serializers.CharField(max_length=100)
+    description = serializers.CharField(required=False)
+    duration = serializers.IntegerField(required=True)
```
LiudmylaKulinchenko, line 9: isn't it required by default?

cinema/settings.py
```diff
@@ -0,0 +1,125 @@
...
+
+INSTALLED_APPS = [
+    "django.contrib.admin",
+    "django.contrib.auth",
+    "django.contrib.contenttypes",
+    "django.contrib.sessions",
+    "django.contrib.messages",
+    "django.contrib.staticfiles",
+    "cinema",
+    "rest_framework",
```
LiudmylaKulinchenko, line 41: Your custom apps should be last in this list

cinema/views.py
```diff
@@ -0,0 +1,38 @@
...
+def movie_list(request):
+    if request.method == "GET":
+        movies = Movie.objects.all()
+        serializer = MovieSerializer(movies, many=True)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+    else:
+        serializer = MovieSerializer(data=request.data)
+        if serializer.is_valid(raise_exception=True):
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_201_CREATED)
```
LiudmylaKulinchenko, line 20: no need to use else here, since you specified available methods in decorator

cinema/views.py
```diff
@@ -0,0 +1,38 @@
...
+        return Response(serializer.data, status=status.HTTP_200_OK)
+    elif request.method == "PUT":
+        serializer = MovieSerializer(movie, data=request.data)
+        if serializer.is_valid(raise_exception=True):
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_200_OK)
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+    elif request.method == "DELETE":
+        movie.delete()
+        return Response(status=status.HTTP_204_NO_CONTENT)
```
LiudmylaKulinchenko, line 38: same here about delete. Also try to be consistent: in your previous method in the same case you used `else`, but here you used `elif`

LiudmylaKulinchenko (changes requested):

cinema/urls.py
```diff
@@ -0,0 +1,30 @@
...
+
+from cinema.views import movie_list, movie_detail
+
+
+app_name = "cinema"
+
+urlpatterns = [
+    path("admin/", admin.site.urls),
+    path("api/cinema/movies/", movie_list, name="cinema"),
+    path("api/cinema/movies/<int:pk>/", movie_detail, name="cinema_detail"),
```
MNDonut: actually you had to create a separate app(name it cinema) and rename the main app. After that you should use api/cinema/ and include function to include urls from that cinema app

cinema/views.py
```diff
@@ -0,0 +1,38 @@
...
+from cinema.serializers import MovieSerializer
+
+
+@api_view(["GET", "POST"])
+def movie_list(request):
+    if request.method == "GET":
+        movies = Movie.objects.all()
+        serializer = MovieSerializer(movies, many=True)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+    else:
```
MNDonut: you can remove else statement because your view supports only GET and POST. If GET returns Response, then it won't reach this part. Otherwise it's POST method

cinema/views.py
```diff
@@ -0,0 +1,38 @@
...
+        serializer = MovieSerializer(data=request.data)
+        if serializer.is_valid(raise_exception=True):
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_201_CREATED)
+    return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```
MNDonut, line 21: you can remove it because your view supports only 2 methods, others are prohibited

cinema/views.py
```diff
@@ -0,0 +1,38 @@
...
+    if request.method == "GET":
+        serializer = MovieSerializer(movie)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+    elif request.method == "PUT":
+        serializer = MovieSerializer(movie, data=request.data)
+        if serializer.is_valid(raise_exception=True):
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_200_OK)
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+    else:
```
MNDonut: you can remove else statement because your view supports only GET, PUT AND DELETE. If GET or PUT returns Response, then it won't reach this part. Otherwise it's DELETE method

MNDonut (changes requested):

fsocie7y (changes requested):
pls remove this files from ur PR) 
image

fsocie7y (approved):
well done ^_^

PR #582, opened 2024-05-24
Solution

Ruxxell54:


myproject/myproject/cinema/models.py
```diff
@@ -0,0 +1,10 @@
+from django.db import models
+
+
+class Movie(models.Model):
+    title = models.TextField(max_length=255, unique=True)
+    description = models.TextField(max_length=255, blank=True, null=True)
```
LiudmylaKulinchenko, line 6: description shouldn't be nullable

myproject/myproject/cinema/serializers.py
```diff
@@ -0,0 +1,18 @@
+from rest_framework import serializers
+from cinema.models import Movie
+
+
+class MovieSerializer(serializers.Serializer):
+    id = serializers.IntegerField(read_only=True)
+    title = serializers.CharField(required=True, max_length=255, unique=True)
+    description = serializers.CharField(required=False, max_length=255)
+    duration = serializers.IntegerField(required=True)
+
```
LiudmylaKulinchenko, line 10: it's required=True by default

myproject/myproject/cinema/views.py
```diff
@@ -0,0 +1,35 @@
...
+        serializer = MovieSerializer(movie)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+    if request.method == "PUT":
+        serializer = MovieSerializer(movie, data=request.data)
+        serializer.is_valid(raise_exception=True)
+        serializer.save()
+        return Response(serializer.data, status=status.HTTP_200_OK)
+    else:
+        movie.delete()
+        return Response(status=status.HTTP_204_NO_CONTENT)
```
LiudmylaKulinchenko, line 35: No need to use else here (you didn't use it in the same case in previous function)

myproject/myproject/cinema_service/settings.py
LiudmylaKulinchenko, line 1: fix quotes everywhere (use `black`)

LiudmylaKulinchenko (changes requested):

YuriiKindrat (approved):
GJ

PR #587, opened 2024-05-26
Solution

SerhiiMar:


cinema/views.py
```diff
@@ -0,0 +1,38 @@
...
+from cinema.serializers import MovieSerializer
+
+
+@api_view(["GET", "POST"])
+def movie_list(request):
+    if request.method == "GET":
+        movies = Movie.objects.all()
+        serializer = MovieSerializer(movies, many=True)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+    else:
```
MNDonut: you can remove else statement because your view supports only GET and POST. If GET returns Response, then it won't reach this part. Otherwise it's POST method

cinema/views.py
```diff
@@ -0,0 +1,38 @@
...
+    if request.method == "GET":
+        serializer = MovieSerializer(movie)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+    elif request.method == "PUT":
+        serializer = MovieSerializer(movie, data=request.data)
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_200_OK)
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+    else:
```
MNDonut: the same case as above

MNDonut (changes requested):

YuriiKindrat (approved):
GJ

PR #607, opened 2024-06-16
added solution to movie api

AlexRozv:


app/settings.py
```diff
@@ -0,0 +1,113 @@
...
+
+INSTALLED_APPS = [
+    "django.contrib.admin",
+    "django.contrib.auth",
+    "django.contrib.contenttypes",
+    "django.contrib.sessions",
+    "django.contrib.messages",
+    "django.contrib.staticfiles",
+    "cinema",
+    "rest_framework"
```
LiudmylaKulinchenko: Your custom apps should be last in this list:
1. default
2. third-party
3. custom

cinema/serializers.py
```diff
@@ -0,0 +1,23 @@
+from rest_framework import serializers
+
+from cinema.models import Movie
+
+
+class MovieSerializer(serializers.Serializer):
+    id = serializers.IntegerField(read_only=True)
+    title = serializers.CharField(max_length=255)
+    description = serializers.CharField(required=False)
```
LiudmylaKulinchenko: Why do you think this field shouldn't be required? Wouldn't it be confusing if you try to buy a movie ticket, for example, and are not able to see any description?

LiudmylaKulinchenko (changes requested):
Add type annotations, please!

LiudmylaKulinchenko (approved):
Looks good 🚀

PR #609, opened 2024-06-17
Solution

Grotski:


.gitignore
```diff
@@ -0,0 +1,10 @@
+.idea/
+.env
+.venv
+env/
+venv/
+ENV/
+env.bak/
+venv.bak/
+**__pycache__
+db.sqlite3
```
vsmutok: add new empty line

cinema/views.py
```diff
@@ -0,0 +1,39 @@
...
+def movie_list(request):
+    if request.method == "GET":
+        movies = Movie.objects.all()
+        serializer = MovieSerializer(movies, many=True)
+        return Response(serializer.data, safe=False, status=status.HTTP_200_OK)
+    else:
+        serialazer = MovieSerializer(data=request.data)
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_201_CREATED)
```
vsmutok: Avoid using an if condition to check if a serializer is valid. Instead, use the raise_exception=True flag when calling serializer.is_valid(). This will automatically raise a ValidationError if the data is invalid, which is then caught by the DRF exception handler to return a 400 Bad Request response.


read this file https://github.com/mate-academy/py-movie-api/blob/master/checklist.md

vsmutok (changes requested):

EdAlekseiev (approved):

PR #620, opened 2024-07-06
Solution

gmuhamax:


YukihiroSM (changes requested):
Remove .idea folder from git, please. 
https://stackoverflow.com/questions/6313126/how-do-i-remove-a-directory-from-a-git-repository

cinema/views.py
```diff
@@ -0,0 +1,43 @@
...
+
+@api_view(["GET", "POST"])
+def movie_list(request) -> Response:
+    if request.method == "GET":
+        serializer = MovieSerializer(Movie.objects.all(), many=True)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+    elif request.method == "POST":
+        serializer = MovieSerializer(data=request.data)
+        if serializer.is_valid():
+            serializer.save()
```

vsmutok (changes requested):

vasylhnatiuk (approved):

PR #635, opened 2024-07-25
Develop

SemenetsSergii:
image
image
image


Arsen-hrynevych (changes requested):
Good job! I really appreciate your effort .  Please remove the unnecessary files.

api/settings.py
```diff
@@ -0,0 +1,134 @@
...
+
+INSTALLED_APPS = [
+    "django.contrib.admin",
+    "django.contrib.auth",
+    "django.contrib.contenttypes",
+    "django.contrib.sessions",
+    "django.contrib.messages",
+    "django.contrib.staticfiles",
+    "cinema",
+    "rest_framework",
```
Arsen-hrynevych, line 43: Are you sure about this order?

Arsen-hrynevych (changes requested):
Good job! I really appreciate your effort . Please review the check list of the task. Please remove cpython files

api/settings.py
SemenetsSergii: yes, of course

Arsen-hrynevych (changes requested):
Забери будь ласка кеш файли, їх не має бути тут

api/settings.py
Arsen-hrynevych: In the INSTALLED_APPS setting of a Django project, the order of apps can be significant, especially for certain types of applications. While it is not strictly required for rest_framework to be listed first, it is often beneficial to place it before your custom apps like cinema. Here are the reasons why you might want to do this:

Middleware and Request Processing Order:

Django processes middleware in the order it is defined in the MIDDLEWARE setting. While INSTALLED_APPS does not directly affect middleware order, having rest_framework listed before custom apps can sometimes prevent conflicts in middleware that depend on the order of app initialization.
Template Loading:

Django looks for templates in the order of apps listed in INSTALLED_APPS. If there are template name conflicts, Django will use the template from the first app that matches. Listing rest_framework before custom apps ensures that its default templates are available before any potential overrides in custom apps.
Signals and App Initialization:

Django initializes apps in the order they are listed. This affects signal registration and app-specific initialization code. rest_framework might register certain signals or perform initialization that should happen before custom apps do their own setup.
Migration Order:

The order in INSTALLED_APPS can affect the order in which migrations are applied. While this usually does not cause issues, having rest_framework listed before custom apps can ensure that its models are migrated first, which can be important if your app depends on any rest_framework models.
Consistency and Convention:

Many Django projects follow the convention of listing third-party apps before custom apps. This improves readability and maintainability, making it clear which apps are third-party and which are custom.
In your specific case, placing rest_framework after cinema may not cause any immediate issues, but following the conventional order can prevent potential conflicts and make the configuration more predictable.

Arsen-hrynevych (approved):
А оновлювати PR хто буде? Я глянув у тебе в репо ти видалив pycache,  проте варто перевіряти що відправляєш на перевірку

SemenetsSergii:
Arsen, thanks for the explanation, now I understand what it is about (I had no idea about it) Thanks for the detailed explanation))) I'm just learning, but I understand that there are things that have bits in their places, clean code)

SemenetsSergii:
> А оновлювати PR хто буде? Я глянув у тебе в репо ти видалив pycache, проте варто перевіряти що відправляєш на перевірку

все розыбрався)))
image

PR #657, opened 2024-08-06
'Solution'

OLENA-KALITSINSKA:


cinema/views.py
```diff
@@ -0,0 +1,36 @@
...
+
+@api_view(["GET", "PUT", "DELETE"])
+def movie_detail(request, pk):
+    movie = get_object_or_404(Movie, pk=pk)
+    if request.method == "GET":
+        serializer = MovieSerializer(movie)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+    if request.method == "PUT":
+        serializer = MovieSerializer(movie, data=request.data)
+        if serializer.is_valid():
```
vasylhnatiuk: use serializer.is_valid(raise_exception=True) here as well

vasylhnatiuk (changes requested):

cinema/views.py
```diff
@@ -0,0 +1,36 @@
...
+def movie_detail(request, pk):
+    movie = get_object_or_404(Movie, pk=pk)
+    if request.method == "GET":
+        serializer = MovieSerializer(movie)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+    if request.method == "PUT":
+        serializer = MovieSerializer(movie, data=request.data)
+        if serializer.is_valid(raise_exception=True):
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_200_OK)
```

vsmutok (changes requested):

vsmutok (approved):
Great!

PR #668, opened 2024-08-16
Implement Movie API

anton-bilous:


cinema/views.py
```diff
@@ -0,0 +1,41 @@
+from django.http import HttpRequest
+from django.shortcuts import get_object_or_404
+from rest_framework import status
+from rest_framework.response import Response
+from rest_framework.decorators import api_view
+
+from .models import Movie
```
Oleksl888: avoid using relative imports

cinema/views.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+@api_view(["GET", "POST"])
+def movies(request: HttpRequest) -> Response:
+    if request.method == "GET":
+        movies = Movie.objects.all()
+        serializer = MovieSerializer(movies, many=True)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+
+    if request.method == "POST":
```
Oleksl888: no need for this condition here, you can just unindent below code because only two methods are accepted and you have return statement above

cinema/views.py
```diff
@@ -0,0 +1,41 @@
...
+from rest_framework import status
+from rest_framework.response import Response
+from rest_framework.decorators import api_view
+
+from .models import Movie
+from .serializers import MovieSerializer
+
+
+@api_view(["GET", "POST"])
+def movies(request: HttpRequest) -> Response:
```
Oleksl888: rename this, so it is easy to understand in other parts of application that this is a view and what it does

cinema/views.py
```diff
@@ -0,0 +1,41 @@
...
+
+    if request.method == "PUT":
+        movie = get_object_or_404(Movie, pk=pk)
+        serializer = MovieSerializer(movie, request.data)
+        serializer.is_valid(raise_exception=True)
+        serializer.save()
+        return Response(serializer.data, status=status.HTTP_200_OK)
+
+    if request.method == "DELETE":
+        get_object_or_404(Movie, pk=pk).delete()
```
Oleksl888: since only these 3 methods are accepted here you don't need last if statement + you can move `movie = get_object_or_404(Movie, pk=pk)` before all if's and then reuse variable in each conditional

.pre-commit-config.yaml
Oleksl888, line 1: well done on adding pre commit hooks :wink:

Oleksl888 (changes requested):

Oleksl888 (approved):
Well done!

PR #670, opened 2024-08-23
Develop

Wallian169:


.gitignore
```diff
@@ -0,0 +1,10 @@
+.idea/
+.vscode/
+*.iml
+.env
+.DS_Store
+venv/
+.pytest_cache/
+**__pycache__/
+*.sqlite3
+.pyc
```
vsmutok: add new empty line

cinema/views.py
```diff
@@ -0,0 +1,43 @@
...
+    if request.method == "PUT":
+        serializer = MovieSerializer(instance=Movie, data=request.data)
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_200_OK)
+        else:
+            return Response(
+                serializer.errors,
+                status=status.HTTP_400_BAD_REQUEST
+            )
```

vsmutok (changes requested):

cinema/views.py
```diff
@@ -0,0 +1,37 @@
...
+from cinema.serializers import MovieSerializer
+
+
+@api_view(["GET", "POST"])
+def cinema_list_view(request: HttpRequest) -> HttpResponse:
+    if request.method == "GET":
+        movies = Movie.objects.all()
+        serializer = MovieSerializer(movies, many=True)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+    else:
```
pavlopro: Please check explicitly, that method is POST

cinema/views.py
```diff
@@ -0,0 +1,37 @@
...
+    movie = get_object_or_404(Movie, pk=pk)
+    if request.method == "GET":
+        serializer = MovieSerializer(movie)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+    if request.method == "PUT":
+        serializer = MovieSerializer(instance=Movie, data=request.data)
+        serializer.is_valid(raise_exception=True)
+        serializer.save()
+        return Response(serializer.data, status=status.HTTP_200_OK)
+    else:
```
pavlopro: Please check explicitly, that method is DELETE

pavlopro (changes requested):

pavlopro (approved):
GJ!

PR #672, opened 2024-08-30
Develop

anastmishchuk:


cinema/views.py
```diff
@@ -0,0 +1,36 @@
...
+def movie_detail(request, pk):
+    movie = get_object_or_404(Movie, pk=pk)
+    if request.method == "GET":
+        serializer = MovieSerializer(movie)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+    elif request.method == "PUT":
+        serializer = MovieSerializer(movie, data=request.data)
+        if serializer.is_valid(raise_exception=True):
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_200_OK)
```

cinema/views.py
```diff
@@ -0,0 +1,36 @@
...
+def movie_list(request):
+    if request.method == "GET":
+        movies = Movie.objects.all()
+        serializer = MovieSerializer(movies, many=True)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+    else:
+        serializer = MovieSerializer(data=request.data)
+        if serializer.is_valid(raise_exception=True):
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_201_CREATED)
```
vsmutok: Please explicitly check for method

vsmutok (changes requested):

cinema/views.py
```diff
@@ -0,0 +1,36 @@
...
+        serializer = MovieSerializer(movie)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+    elif request.method == "PUT":
+        serializer = MovieSerializer(movie, data=request.data)
+        serializer.is_valid(raise_exception=True)
+        serializer.save()
+        return Response(serializer.data, status=status.HTTP_200_OK)
+    else:
+        movie.delete()
+        return Response(status=status.HTTP_204_NO_CONTENT)
```
vsmutok: Please check explicitly, that method is DELETE

vsmutok (changes requested):

.gitignore
```diff
@@ -0,0 +1,10 @@
+.idea/
+.vscode/
+*.iml
+.env
+.DS_Store
+venv/
+.pytest_cache/
+**__pycache__/
+*.sqlite3
+.pyc
```
vasylhnatiuk: add newline

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #677, opened 2024-09-05
Implement solution

dimak20:


cinema/admin.py
```diff
@@ -0,0 +1,5 @@
+from django.contrib import admin
+
+from .models import Movie
```
Oleksl888: avoid using relative imports

cinema/views.py
```diff
@@ -0,0 +1,36 @@
...
+from rest_framework.decorators import api_view
+from rest_framework.generics import get_object_or_404
+from rest_framework.response import Response
+
+from cinema.models import Movie
+from cinema.serializer import MovieSerializer
+
+
+@api_view(["GET", "POST"])
+def movie_list(request) -> Response:
```
Oleksl888: type hint for request?

cinema/views.py
```diff
@@ -0,0 +1,36 @@
...
+    if request.method == "GET":
+        movies = Movie.objects.all()
+        serializer = MovieSerializer(movies, many=True)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+    else:
```
Oleksl888: no need for else here, you can just unindent below code because only two methods are accepted and you have return statement above

cinema/views.py
Oleksl888: i would rename this view to reflect what it is doing (is not only list)

cinema/views.py
```diff
@@ -0,0 +1,36 @@
...
+        serializer = MovieSerializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+        serializer.save()
+        return Response(serializer.data, status=status.HTTP_201_CREATED)
+
+
+@api_view(["GET", "PUT", "DELETE"])
+def movie_detail(request, pk) -> Response:
```
Oleksl888: type hints for request and pk

cinema/views.py
```diff
@@ -0,0 +1,36 @@
...
+    movie = get_object_or_404(Movie, pk=pk)
+    if request.method == "GET":
+        serializer = MovieSerializer(movie)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+    elif request.method == "PUT":
+        serializer = MovieSerializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+        serializer.save()
+        return Response(serializer.data, status=status.HTTP_200_OK)
+    else:
```
Oleksl888: no need for this else

cinema/views.py
Oleksl888: change name here also to reflect that it can also update and delete

cinema_service/settings.py
```diff
@@ -0,0 +1,125 @@
...
+
+INSTALLED_APPS = [
+    "django.contrib.admin",
+    "django.contrib.auth",
+    "django.contrib.contenttypes",
+    "django.contrib.sessions",
+    "django.contrib.messages",
+    "django.contrib.staticfiles",
+    "cinema",
+    "rest_framework",
```
Oleksl888, line 38: build in django modules and external packages must come before your apps. Here and in middleware order is important sometimes

Oleksl888 (changes requested):

cinema/views.py
```diff
@@ -0,0 +1,37 @@
...
+        return Response(serializer.data, status=status.HTTP_200_OK)
+
+    serializer = MovieSerializer(data=request.data)
+    serializer.is_valid(raise_exception=True)
+    serializer.save()
+    return Response(serializer.data, status=status.HTTP_201_CREATED)
+
+
+@api_view(["GET", "PUT", "DELETE"])
+def movie_retrieve_patch_delete(request: HttpRequest, pk: int) -> Response:
```
Oleksl888, line 25: patch is not implemented here but it's not a big issue

Oleksl888 (approved):
Well done! Hope you can play with docker now

PR #684, opened 2024-09-20
implemented API with CRUD operations for the Movie model

MasakDirt:


cinema/models.py
```diff
@@ -0,0 +1,10 @@
+from django.db import models
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField(null=True, blank=True)
```
dimak20: Was it mentioned to make this field non-required?

cinema/serializers.py
```diff
@@ -0,0 +1,24 @@
+from rest_framework import serializers
+
+from cinema.models import Movie
+
+
+class MovieSerializer(serializers.Serializer):
+    id = serializers.IntegerField(read_only=True)
+    title = serializers.CharField(max_length=255, required=True)
+    description = serializers.CharField(required=False, allow_blank=True)
```
dimak20: Also requiements to this field

cinema/views.py
```diff
@@ -0,0 +1,37 @@
...
+from cinema.serializers import MovieSerializer
+
+
+@api_view(["GET", "POST"])
+def movie_list(request):
+    if request.method == "GET":
+        movies = Movie.objects.all()
+        serializer = MovieSerializer(movies, many=True)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+    else:
```
dimak20: You don't need to write an else statement because you only have two allowed methods.

cinema/views.py
```diff
@@ -0,0 +1,37 @@
...
+from rest_framework.decorators import api_view
+from rest_framework.generics import get_object_or_404
+from rest_framework.response import Response
+
+from cinema.models import Movie
+from cinema.serializers import MovieSerializer
+
+
+@api_view(["GET", "POST"])
+def movie_list(request):
```
dimak20: I'd rename this in order to reflect actions that this func provides

dimak20: Type for request and also write what this function returns

cinema/views.py
```diff
@@ -0,0 +1,37 @@
...
+        return Response(serializer.data, status=status.HTTP_200_OK)
+    else:
+        serializer = MovieSerializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+        serializer.save()
+        return Response(serializer.data, status=status.HTTP_201_CREATED)
+
+
+@api_view(["GET", "PUT", "DELETE"])
+def movie_detail(request, pk):
```
dimak20: The same

cinema/views.py
```diff
@@ -0,0 +1,37 @@
...
+
+    if request.method == "GET":
+        serializer = MovieSerializer(movie)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+    elif request.method == "PUT":
+        serializer = MovieSerializer(movie, data=request.data)
+        serializer.is_valid(raise_exception=True)
+        serializer.save()
+        return Response(serializer.data, status=status.HTTP_200_OK)
+    else:
```
dimak20: No need to write else here also (there is only 1 allowed method after GET and PUT)

dimak20 (changes requested):

cinema/models.py
MasakDirt: I decided by myself to do it, non-required because about fields mentioned almost nothing)

Arsen-hrynevych (approved):
Thanks

PR #685, opened 2024-09-20
'Solution'

oleksashcherbakov:


oleksashcherbakov:
image
image
image
image

vsmutok (changes requested):
Please remove `.idea` from PR

.gitignore
```diff
@@ -0,0 +1,12 @@
...
+*.iml
+.env
+.DS_Store
+.venv
+.pytest_cache/
+**__pycache__/
+db.sqlite3
+.sqlite
+.db3
+.pyc
```
pavlopro: add blank line

pavlopro (changes requested):
Where's your all other files?

cinema/views.py
```diff
@@ -0,0 +1,37 @@
...
+def movies_list(request) -> HttpResponse:
+    if request.method == "GET":
+        movies = Movie.objects.all()
+        serializer = MovieSerializer(movies, many=True)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+    else:
+        serializer = MovieSerializer(data=request.data)
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_201_CREATED)
```

cinema/views.py
```diff
@@ -0,0 +1,37 @@
...
+    if request.method == "GET":
+        serializer = MovieSerializer(movie)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+    elif request.method == "PUT":
+        serializer = MovieSerializer(movie, data=request.data)
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_200_OK)
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+    else:
```
vsmutok: Check method explicitly before performing an operation, make sure that it is DELETE before returning a response

vsmutok (changes requested):

cinema/views.py
```diff
@@ -0,0 +1,35 @@
...
+from cinema.models import Movie
+from cinema.serializers import MovieSerializer
+
+@api_view(["GET", "POST"])
+def movies_list(request) -> HttpResponse:
+    if request.method == "GET":
+        movies = Movie.objects.all()
+        serializer = MovieSerializer(movies, many=True)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+    else:
```
pavlopro: Check method explicitly before performing an operation, make sure that it is POST before returning a response

user/admin.py
```diff
@@ -0,0 +1,6 @@
+from django.contrib import admin
```
pavlopro: Why do you have this user app in PR?
Remove it

pavlopro (changes requested):

cinema/views.py
pavlopro: how about this one?

pavlopro (changes requested):

pavlopro (approved):
GJ!

PR #688, opened 2024-09-23
Solution

BitRxp:


cinema/urls.py
```diff
@@ -0,0 +1,9 @@
+from django.urls import path
+from cinema.views import movie_detail, movie_list
+
+app_name = "cinema"
+
+urlpatterns = [
+    path("movies/", movie_list, name="movie_list"),
+    path("movies/<int:pk>", movie_detail, name="movie_detail"),
```
vsmutok:
```suggestion
    path("movies/<int:pk>/", movie_detail, name="movie_detail"),
```

cinema/views.py
```diff
@@ -0,0 +1,39 @@
...
+            return Response(serializer.data, status=status.HTTP_201_CREATED)
+    return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+
+
+@api_view(["GET", "PUT", "DELETE"])
+def movie_detail(request, movie_id):
+    movie = get_object_or_404(Movie, pk=movie_id)
+    if request.method == "GET":
+        serializer = MovieSerializer(movie)
+        return Response(serializer.data, status=status.HTTP_200_OK)
```
vsmutok, line 29: Avoid using an `if` condition to check if a serializer is valid. Instead, use the `raise_exception=True` flag when calling `serializer.is_valid()`. This will automatically raise a `ValidationError` if the data is invalid, which is then caught by the DRF exception handler to return a `400 Bad Request` response.

cinema/views.py
```diff
@@ -0,0 +1,39 @@
...
+    elif request.method == "PUT":
+        movie = get_object_or_404(Movie, pk=movie_id)
+        serializer = MovieSerializer(movie, data=request.data)
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_200_OK)
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+    else:
+        movie.delete()
+        return Response(status=status.HTTP_204_NO_CONTENT)
```
vsmutok: Check method explicitly before performing an operation, make sure that it is DELETE before returning a response

vsmutok (changes requested):

vasylhnatiuk (approved):

PR #690, opened 2024-09-24
Developer

AndrZhuk:


.gitignore
```diff
@@ -0,0 +1,4 @@
+.idea
+.venv
+db.sqlite3
+.pyc
```
pavlopro: add blank line

cinema/views.py
```diff
@@ -0,0 +1,39 @@
...
+
+@api_view(["GET", "POST"])
+def movies_list(request):
+    if request.method == "GET":
+        movies = Movie.objects.all()
+        serializer = MovieSerializer(movies, many=True)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+    elif request.method == "POST":
+        serializer = MovieSerializer(data=request.data)
+        if serializer.is_valid():
```
pavlopro: Avoid using an if condition to check if a serializer is valid. Instead, use the `raise_exception=True` flag when calling `serializer.is_valid(`). This will automatically raise a ValidationError if the data is invalid, which is then caught by the DRF exception handler to return a 400 Bad Request response.

cinema/views.py
```diff
@@ -0,0 +1,39 @@
...
+    if request.method == "GET":
+        serializer = MovieSerializer(movie)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+    elif request.method == "PUT":
+        serializer = MovieSerializer(movie, data=request.data)
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_200_OK)
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+    else:
```
pavlopro: Check method explicitly before performing an operation, make sure that it is `DELETE` before returning a response

pavlopro (changes requested):

cinema/views.py
```diff
@@ -0,0 +1,37 @@
...
+
+@api_view(["GET", "PUT", "DELETE"])
+def movie_detail(request, pk):
+    movie = get_object_or_404(Movie, pk=pk)
+    if request.method == "GET":
+        serializer = MovieSerializer(movie)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+    elif request.method == "PUT":
+        serializer = MovieSerializer(movie, data=request.data)
+        if serializer.is_valid(raise_exception=True):
```
pavlopro: In case using `raise_exception=True`, no need to have this `if` operator here. Remove it

cinema/views.py
```diff
@@ -0,0 +1,37 @@
...
+
+@api_view(["GET", "POST"])
+def movies_list(request):
+    if request.method == "GET":
+        movies = Movie.objects.all()
+        serializer = MovieSerializer(movies, many=True)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+    elif request.method == "POST":
+        serializer = MovieSerializer(data=request.data)
+        if serializer.is_valid(raise_exception=True):
```
pavlopro: In case using `raise_exception=True`, no need to have this `if` operator here. Remove it

pavlopro (changes requested):

pavlopro (approved):
GJ!

PR #704, opened 2024-09-26
Develop

vladislav-tsybuliak1:


cinema/serializers.py
```diff
@@ -0,0 +1,23 @@
+from rest_framework import serializers
+
+from cinema.models import Movie
+
+
+class MovieSerializer(serializers.Serializer):
+    id = serializers.IntegerField(read_only=True)
+    title = serializers.CharField(max_length=255, required=True)
+    description = serializers.CharField(required=False)
```
mnezbrytska, line 9: A serializer field is required by default. (info in Check list), and in models you have TextField

mnezbrytska (changes requested):

mnezbrytska (approved):

mnezbrytska (approved):

cinema/serializers.py
vladislav-tsybuliak1: I aggree that required is reduntant here.
In Django Rest serializers.CharField corresponds to django.db.models.fields.CharField or django.db.models.fields.TextField. That's why I've used it here. Serializers don't have "Text field".

Arsen-hrynevych (approved):
Thanks

PR #705, opened 2024-09-26
Solution Politov

Alexandr-Politov:


cinema/admin.py
```diff
@@ -0,0 +1,3 @@
+from django.contrib import admin
+
+# Register your models here.
```
pashawarganov, line 3: I think it`s a good practice to register your models for admin page

cinema/views.py
```diff
@@ -0,0 +1,25 @@
...
+        serializer = MovieSerializer(movies, many=True)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+    elif request.method == "POST":
+        serializer = MovieSerializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+        serializer.save()
+        return Response(serializer.data, status=status.HTTP_201_CREATED)
+
+
+@api_view
```
pashawarganov: You need implement method for get, put and delete single movie

pashawarganov (changes requested):
You also need implement serializers

Arsen-hrynevych (approved):
Thanks

PR #706, opened 2024-09-26
solution py-movie-api

Maksym-Turenko:


cinema/serializers.py
```diff
@@ -0,0 +1,22 @@
...
+
+class MovieSerializer(serializers.Serializer):
+    id = serializers.IntegerField(read_only=True)
+    description = serializers.CharField()
+    duration = serializers.IntegerField()
+
+    def create(self, validated_data: dict) -> Movie:
+        return Movie.objects.create(**validated_data)
+
+    def update(self, instance, validated_data):
```
AndriiHamasa: In the create method, you used annotations, but in the update method, you didn't. You should either use them everywhere or not use them at all.

cinema/views.py
```diff
@@ -0,0 +1,44 @@
...
+@api_view(["GET", "POST"])
+def movie_list(request: Request) -> Response:
+    if request.method == "GET":
+        movies = Movie.objects.all()
+        serializer = MovieSerializer(movies, many=True)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+
+    elif request.method == "POST":
+        serializer = MovieSerializer(data=request.data)
+        if serializer.is_valid():
```
AndriiHamasa: Refer to the checklist.md file, where it's recommended to use:
``serializer.is_valid(raise_exception=True)``
instead of using an "if" statement. Please apply this fix in the detail view as well.

AndriiHamasa (changes requested):
Additionally, I'm not entirely sure, but it seems like there might be a missing empty line at the end of the files.

Arsen-hrynevych (changes requested):
Remove db.sqlite, idea, __pycache__ from PR

oliasymak24 (approved):

PR #707, opened 2024-09-26
"Solution"

mnezbrytska:


cinema/serializers.py
```diff
@@ -0,0 +1,27 @@
...
+            "description",
+            instance.description
+        )
+        instance.duration = validated_data.get("duration", instance.duration)
+        instance.save()
+        return instance
+
+    class Meta:
+        model = Movie
+        fields = "__all__"
```
olena4reunited: Using "__all__" is convenient overall, but it would be safer to list all the fields to avoid exposing unintended fields if the model changes

cinema/views.py
```diff
@@ -0,0 +1,35 @@
...
+def movie_list(request):
+    if request.method == "GET":
+        movies = Movie.objects.all()
+        serializer = MovieSerializer(movies, many=True)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+    else:
+        serializer = MovieSerializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+        serializer.save()
+        return Response(serializer.data, status=status.HTTP_201_CREATED)
```
olena4reunited, line 19: Non valid data could be send in POST request. You should handle this case.

cinema/views.py
```diff
@@ -0,0 +1,35 @@
...
+def movie_detail(request, pk):
+    movie = Movie.objects.get(pk=pk)
+    if request.method == "GET":
+        serializer = MovieSerializer(movie)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+    elif request.method == "PUT":
+        serializer = MovieSerializer(movie, data=request.data)
+        serializer.is_valid(raise_exception=True)
+        serializer.save()
+        return Response(serializer.data, status=status.HTTP_200_OK)
```
olena4reunited, line 32: Non valid data could be send in PUT request. You should handle this case.

olena4reunited (changes requested):
Good job, but don't forget about handling non valid data

cinema/views.py
mnezbrytska: Mayby, I didn't understand you correct, but in checklist we have: "Avoid using an if condition to check if a serializer is valid. Instead, use the raise_exception=True flag when calling serializer.is_valid(). This will automatically raise a ValidationError if the data is invalid, which is then caught by the DRF exception handler to return a 400 Bad Request response."

olena4reunited: Good notice! My bad.

olena4reunited (approved):

cinema/serializers.py
```diff
@@ -24,4 +24,4 @@ def update(self, instance, validated_data):
 
     class Meta:
         model = Movie
-        fields = "__all__"
+        fields = "id", "title", "description", "duration"
```
olena4reunited, line 27: Put your fields names into list

olena4reunited (changes requested):

Arsen-hrynevych (approved):
Thanks

PR #721, opened 2024-10-01
Solution

jyjuk:
01 10 2024_09 13 43_REC
01 10 2024_09 15 40_REC
01 10 2024_09 16 28_REC
01 10 2024_09 16 55_REC


cinema/views.py
```diff
@@ -0,0 +1,36 @@
...
+    movie = get_object_or_404(Movie, pk=pk)
+    if request.method == "GET":
+        serializer = MovieSerializer(movie)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+    elif request.method == "PUT":
+        serializer = MovieSerializer(movie, data=request.data)
+        serializer.is_valid(raise_exception=True)
+        serializer.save()
+        return Response(serializer.data, status=status.HTTP_200_OK)
+    else:
```
Dimosphen1, line 34: In some cases it is better to explicitly check that the method is DELETE and do not default to it using `else`

Dimosphen1 (approved):
LGTM, well done! One comment is left for consideration

cinema/views.py
jyjuk: Ok

PR #750, opened 2024-11-25
Movie API solutions

SviatoslavTataryn:


pavlopro (changes requested):
remove .idea and __pycache__ folders from PR

PR #768, opened 2024-12-06
movie-app solution

94nj111:


cinema/urls.py
```diff
@@ -0,0 +1,10 @@
+from django.urls import path
+
+from cinema.views import movie_list, movie_detail
+
+app_name = "cinema"
+
+urlpatterns = [
+    path("movies/", movie_list, name="bus_list"),
+    path("movies/<int:pk>/", movie_detail, name="bus_detail"),
```
skyfoxwork: try to use something like: name="movies-list" and name="movie-detail" instead of name="bus_list", name="bus_detail".

94nj111: Sorry🤣

cinema/serializers.py
```diff
@@ -0,0 +1,9 @@
+from rest_framework import serializers
+
+from cinema.models import Movie
+
+
+class MovieSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Movie
+        fields = ["id", "title", "description", "duration"]
```
skyfoxwork, line 9: interesting  solution. I don't know that.

skyfoxwork (approved):
Good job!

cinema/urls.py
skyfoxwork: Don't worry it is minor error. I do that all time :)

PR #777, opened 2024-12-08
solution

VolodymyrSemchysyn:


Le0n-K (changes requested):
Where is your files with models and views?👀

PR #778, opened 2024-12-08
Solution

Le0n-K:


yulia-pl (changes requested):
You should delete this commit and manually add only the necessary files. Here's an example:
git add cinema/serializers.py
git add cinema/urls.py
...
And so on. Manually add only what you need.

PR #781, opened 2024-12-08
Solution

Le0n-K:


cinema/views.py
```diff
@@ -0,0 +1,37 @@
...
+    if request.method == "GET":
+        movies = Movie.objects.all()
+        serializer = MovieSerializer(movies, many=True)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+    else:
+        serializer = MovieSerializer(data=request.data)
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_201_CREATED)
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```
dxrrkwm: avoid using `if is_valid()`

dxrrkwm (changes requested):
nice! check comment :)

dxrrkwm (changes requested):
fill .gitignore and simplify serializer to inherit from ModelSerializer, description should be TextField

cinema/urls.py
```diff
@@ -0,0 +1,9 @@
+from django.urls import path
+from .views import movie_list, movie_detail
+
+app_name = "cinema"
+
+urlpatterns = [
+    path("movie/", movie_list, name="movie_list"),
+    path("movie/<int:pk>/", movie_detail, name="movie_detail"),
```
dxrrkwm: urls should be in kebab case

dxrrkwm (changes requested):

dxrrkwm (commented):
nice, now you can simplify serializer with only using class Meta

PR #789, opened 2024-12-09
Solution

Bondzik-S:


cinema/views.py
```diff
@@ -0,0 +1,37 @@
...
+@api_view(["GET", "POST"])
+def movies_list(request):
+    if request.method == "GET":
+        movies = Movie.objects.all()
+        serializer = CinemaSerializer(movies, many=True)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+    else:
+        serializer = CinemaSerializer(data=request.data)
+        if serializer.is_valid():
+            serializer.save()
```
sumseer: Although it works, Code Style in checklist.md specifically asks us the following:
1. Avoid using an `if` condition to check if a serializer is valid. Instead, use the `raise_exception=True` flag when calling `serializer.is_valid()`. This will automatically raise a `ValidationError` if the data is invalid, which is then caught by the DRF exception handler to return a `400 Bad Request` response.

sumseer (changes requested):

sumseer (approved):
Looks good now!

PR #806, opened 2025-01-18
Solution

zave52:


cinema/admin.py
vsmutok, line 1: Consider registering your models with the Django admin site. You can do this by importing your models and using `admin.site.register(ModelName)` for each model you want to manage through the admin interface.

vsmutok (changes requested):

vsmutok (approved):
Good Job!

PR #821, opened 2025-02-06
Solution

dottfmar:


cinema/views.py
```diff
@@ -0,0 +1,40 @@
...
+
+@api_view(["GET", "POST"])
+def movie_list(request: Request) -> Response:
+    if request.method == "GET":
+        movies = Movie.objects.all()
+        serializer = MovieSerializer(movies, many=True)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+
+    else:
+        serializer = MovieSerializer(data=request.data, many=True)
```
4ikkamony: here's a little problem, if you POST a single JSON obj, nothing will be created.
A list of json's will work, tho

IDK how to deal with that properly, I personaly check if request.data is a list and set many=True or False
Maybe there's a better way

4ikkamony (approved):
Nice work!

Apart from that little comment, perfect!

PR #824, opened 2025-02-07
Solution

VovaGld:


cinema/serializers.py
```diff
@@ -0,0 +1,19 @@
+from rest_framework import serializers
+from cinema.models import Movie
+
+
+class MovieSerializer(serializers.Serializer):
+    id = serializers.IntegerField(read_only=True)
+    title = serializers.CharField(max_length=100, required=True)
```
taraskhom, line 7: "required=True" isn't necessary because it's true by default

cinema/serializers.py
```diff
@@ -0,0 +1,19 @@
...
+    description = serializers.CharField(required=False)
+    duration = serializers.IntegerField(required=True, min_value=0)
```
taraskhom, line 9: "required=True" isn't necessary because it's true by default

taraskhom (approved):
Good job!

PR #827, opened 2025-02-07
Solution

julia4406:


Sichkaridze (approved):
You accidentally pushed '__pycache__'.
And interesting part, that i did it too.

I don't know why, but it wasn't shown in my git status.
I think same happened to you.

PR #828, opened 2025-02-07
solution

Sichkaridze:


Sichkaridze:
sorry, I accidentally pushed __pycache__, because for I don't know what reason it was not visible in git status

cinema/views.py
```diff
@@ -0,0 +1,38 @@
...
+
+@api_view(["GET", "POST"])
+def movie_list(request):
+    if request.method == "GET":
+        movies = Movie.objects.all()
+        serializer = MovieSerializer(movies, many=True)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+    else:
+        serializer = MovieSerializer(data=request.data)
+        if serializer.is_valid():
```
DanilTrukhanov: In the checklist file, there is a paragraph that says "Avoid using an if condition to check if a serializer is valid."

DanilTrukhanov (approved):
well done!

cinema/views.py
Sichkaridze: Ooo, yeah<) You are right! Thanks =)

PR #832, opened 2025-02-08
Solution movie-api

Bleit1:


.gitignore
```diff
@@ -0,0 +1,7 @@
+venv
+.idea
+db.sqlite3
+__pycache__
+cinema_project/__pycache__/settings.cpython-312.pyc
+cinema_project/__pycache__/urls.cpython-312.pyc
```
Dogober, line 6: you can use such string: `**__pycache__/`
It will help you to avoid repeating yourself

movie_api_cinema/movie_api_cinema/urls.py
```diff
@@ -0,0 +1,7 @@
+from django.contrib import admin
+from django.urls import path, include
+
+urlpatterns = [
+    path("admin/", admin.site.urls),
+    path("api/cinema/", include("cinema.urls", namespace="cinema")),
```
Dogober, line 6: here you can add one more part for your whole path like -> ` path("api/v1/cinema/", include("cinema.urls", namespace="cinema")),`

Dogober (changes requested):
You need to implement relevant views like movie_list_view and movie_detail_view and add relevant pathes for processing requests methods

PR #833, opened 2025-02-08
Solution

Dogober:


cinema/models.py
```diff
@@ -0,0 +1,13 @@
+from django.db import models
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255, unique=True)
```
taraskhom: It's not necessary that title needs to be unique because in real world we have a lot of movies with same title

cinema/models.py
```diff
@@ -0,0 +1,13 @@
...
+    description = models.TextField(null=True, blank=True)
+    duration = models.IntegerField()
+
+    class Meta:
+        verbose_name_plural = "Movies"
```
taraskhom: It will be "movies" in plural without defining class Meta

taraskhom (approved):
Good job!

PR #834, opened 2025-02-08
Solution

bodiakof:


cinema/urls.py
```diff
@@ -0,0 +1,7 @@
+from django.urls import path
+from cinema.views import movie_list, movie_detail
+
```
ruslanyemets, line 3: You should add app_name here

app/urls.py
```diff
@@ -0,0 +1,24 @@
...
+    1. Import the include() function: from django.urls import include, path
+    2. Add a URL to urlpatterns:  path('blog/', include('blog.urls'))
+"""
+
+from django.contrib import admin
+from django.urls import path, include
+
+urlpatterns = [
+    path("admin/", admin.site.urls),
+    path("api/cinema/", include("cinema.urls")),
```
ruslanyemets: You should add namespace to include

cinema/serializers.py
```diff
@@ -0,0 +1,8 @@
+from rest_framework import serializers
+from .models import Movie
+
+
+class MovieSerializer(serializers.ModelSerializer):
```
ruslanyemets, line 5: You should implement serializer for movies with selializer fields (id, title, description, duration) and create, update methods

ruslanyemets (changes requested):
You should implement class MovieSerializer and fix some minor issues in your project. Moreover remove from git repository .idea and __pycache__ directories

cinema/serializers.py
bodiakof: In the task "Implement a serializer for this model." there is no specific condition to use Serializer (instead of ModelSerializer as I did). ModelSerializer simply allows skipping all that boilerplate code that we gotta write with Serializer and handles it all by itself :)

ruslanyemets: Ok, maybe you're right :)

ruslanyemets (approved):
Good job!

PR #840, opened 2025-02-10
Solution

MindDevastation:


cinema/models.py
```diff
@@ -0,0 +1,9 @@
+from django.db import models
+
+class Movie(models.Model):
+    title = models.CharField(max_length=200)
```
Nattalli: suggest you use `2^n-1` value in CharField max_lwngth

cinema/serializers.py
```diff
@@ -0,0 +1,7 @@
+from rest_framework import serializers
+from .models import Movie
```
Nattalli: use absolute imports

cinema/urls.py
```diff
@@ -0,0 +1,12 @@
+from django.urls import path
+from . import views
+
+urlpatterns = [
+    path("api/movies/", views.movie_list, name="movie-list-api"),
```
Nattalli: why do you need `-api` in name? 
just `movie-list` would be better

Nattalli (changes requested):

Nattalli (approved):
ok, one more - make migrations and push them

PR #892, opened 2025-04-08
solution

Bombardier256:


Bombardier256:
@mateacademy-ai-mentor Test are not required and at this moment serializers.Serializer is the way this project should be coded

PR #896, opened 2025-04-09
Dev

Sotnikov-100:


.idea/.gitignore
Y-Havryliv, line 1: delete `.idea` from the PR

cinema/views.py
```diff
@@ -0,0 +1,43 @@
...
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_201_CREATED)
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+
+
+@api_view(["GET", "PUT", "DELETE"])
+def movie_detail(request, pk):
+    try:
+        movie = Movie.objects.get(pk=pk)
```
Y-Havryliv: you can use `get_object_or_404`

Y-Havryliv (changes requested):

cinema/views.py
Sotnikov-100: You're absolutely right!

PR #910, opened 2025-04-21
Add initial Django project structure with Movie model and API endpoints

iSevenpwnz:


cinema/models.py
```diff
@@ -0,0 +1,10 @@
+from django.db import models
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=100)
+    description = models.TextField()
+    duration = models.IntegerField()
```
PrimeGlorious: Better **-->**  **duration = models.PositiveIntegerField()**

cinema/serializers.py
```diff
@@ -0,0 +1,25 @@
+from rest_framework import serializers
+
+from cinema.models import Movie
+
+
+class MovieSerializer(serializers.ModelSerializer):
+    id = serializers.IntegerField(read_only=True)
+    title = serializers.CharField(required=True, max_length=100)
+    description = serializers.CharField(required=True)
+    duration = serializers.IntegerField(required=False)
```
PrimeGlorious: If model field is **PositiveIntegerField** you can use validator:
**validators=[MinValueValidator(1)]**

PrimeGlorious (approved):

PR #912, opened 2025-04-22
start django project, add cinema app and core

BronkstoneBro:


cinema/serializers.py
```diff
@@ -0,0 +1,25 @@
...
+        instance.description = validated_data.get(
+            "description",
+            instance.description
+        )
+        instance.duration = validated_data.get(
+            "duration",
+            instance.duration
+        )
+        instance.save()
+        return instance
```
cemen2009: would be good practice to add a Meta class. I got problem with putting data without it

cemen2009 (approved):

PR #919, opened 2025-04-22
Solution

Oneyura:


DHlomozda (approved):
Good job. Do not forgert about empty line after end of file

PR #924, opened 2025-04-23
Solution py-movie-api

Roksolana-K:


yarzhuravlov (approved):
It would be worth removing .idea and __pycache__
Why is readme.md in .gitignore?

PR #929, opened 2025-04-26
movie api

Maaax256:


Nikita-pyth (changes requested):
Hi, i'm just a bit confused. You do not have urls.py in cinema app and also didn't include it in urls.py in movie_api. Otherwise, the project seems solid, I'm just confused how it can even start up with this setup

Nikita-pyth (approved):
💪

PR #930, opened 2025-04-26
Cinema innit

Tolik-code:


AnyoneClown (changes requested):
Delete all `__pycache__` files and add them to `.gitignore`

PR #932, opened 2025-04-28
'solution'

ntkachenko21:


cinema/tests.py
```diff
@@ -0,0 +1,3 @@
+from django.test import TestCase
+
+# Create your tests here.
```
Boyok1337: If u don`t need it, delete it

Boyok1337 (changes requested):
Remove .idea and pycache

Boyok1337 (approved):
GJ!

PR #1032, opened 2025-09-01
Implemented basic serializer for Movie model

Kostenikov:


MakovM (approved):

Kostenikov:
@mateacademy-ai-mentor i don't need any changes, movie can be successfully created when description is not provided. That's why required=False is defined in serializer description field.

PR #1036, opened 2025-09-01
Solution

bohdankolisnyk04:


DenisPrudius (changes requested):
You must 
Create api views for the following endpoints:

    GET api/cinema/movies/ - should return a list of the all movies
    GET api/cinema/movies/<pk>/ - should return a movie with given id
    POST api/cinema/movies/ - should create a new movie based on passed data
    PUT api/cinema/movies/<pk>/ - should update the movie with given id based on passed data
    DELETE api/cinema/movies/<pk>/ - should delete the movie with given id

PR #1114, opened 2025-12-24
cinema app created, added serializers, created views, gitignore

AlexMishKo:


cinema/models.py
```diff
@@ -0,0 +1,13 @@
+from django.db import models
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
```
Guillelmina8, line 6: Maybe it is better to use models.CharField for description field, so it will be the same as in MovieSerializer.

Guillelmina8 (approved):
Excellent work!

PR #1118, opened 2025-12-24
solution

leemonch1k-space:


Iventyk (approved):
Great!

leemonch1k-space:
@mateacademy-ai-mentor

Your comment "The Movie model uses PositiveIntegerField for duration. To ensure data is validated correctly at the API level before hitting the database, you should use serializers.PositiveIntegerField here. This will prevent invalid (negative) values from being accepted by the API"

Django rest framework does not have "PositiveIntegerField" attribute, but according to your comment, I added "serializers.IntegerField(min_value=0)" which works the same.

PR #1119, opened 2025-12-24
Py movie api

Ann-Riabykina:


VadymVynokur (approved):
we shuld use func base in this task, and classes is next topic, but still
it works so
good job

PR #1127, opened 2025-12-28
Create movie api

SofaD8:


cinema/models.py
```diff
@@ -0,0 +1,10 @@
+from django.db import models
+
+
+class Movie(models.Model):
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    duration = models.IntegerField()
+
+    def __str__(self):
+        return self.title
```
Guillelmina8, line 10: I guess you haven't added all updated files to the PR. I don't see all models, serializers, urls and views. Everything looks like the start project after forking and cloning. Maybe you should use git add .

Guillelmina8 (approved):
Update your changes and I believe that everything is great there!

PR #1197, opened 2026-04-22
Solution

viloker:


cinema/admin.py
```diff
@@ -0,0 +1,3 @@
+from django.contrib import admin
+
+# Register your models here.
```
vsmutok: Remove or register models import: leaving only comment and unused import is fine for admin, but checklist requires removing check/debug comments before commit. Line has placeholder comment - delete or register your Movie model in admin if intended.

vsmutok (changes requested):

Boyok1337 (approved):
LGMT!

PR #1299, opened 2026-08-02
Create a Django project and a cinema app inside it.

JacksonMarch:
Implement a serializer for this model.
Create api views for the following endpoints.


.idea/.gitignore
vsmutok, line 1: Please remove `.idea `folder from the PR

cinema/views.py
```diff
@@ -0,0 +1,36 @@
...
+        serializer = MovieSerializer(movie)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+    elif request.method == "PUT":
+        serializer = MovieSerializer(movie, data=request.data)
+        serializer.is_valid(raise_exception=True)
+        serializer.save()
+        return Response(serializer.data, status=status.HTTP_200_OK)
+    else:
+        movie.delete()
+        return Response(status=status.HTTP_204_NO_CONTENT)
```
vsmutok, line 36: It would be better to explicitly check request.method == "DELETE" instead of handling it in the else branch.

vsmutok (changes requested):

PR #1310, opened 2026-08-13
Complete

Rostyk-D:


.idea/inspectionProfiles/profiles_settings.xml
Sn1kls, line 1: `.idea/` is committed. It's in `.gitignore`, but it was committed before that, so it's still tracked. Remove it: `git rm -r --cached .idea` and commit. (Checklist: don't push `.idea`)

requirements.txt
Sn1kls, line 1: file is saved as UTF-16 (a Windows PowerShell artifact). This can break `pip install -r requirements.txt`. Re-save it as UTF-8, e.g. `pip freeze | Out-File -Encoding utf8 requirements.txt`

Sn1kls (changes requested):

Sn1kls (approved):
Approved, good work! Check the inline comment for a future improvement.

settings/settings.py
```diff
@@ -0,0 +1,129 @@
...
+# Quick-start development settings - unsuitable for production
+# See https://docs.djangoproject.com/en/6.1/howto/deployment/checklist/
+
+# SECURITY WARNING: keep the secret key used in production secret!
+SECRET_KEY = "django-insecure-f40$+or!_gqnsf3hc$__$gf=wg=(5g6)o=aa$*(sk%gki5_o_0"
+
+# SECURITY WARNING: don't run with debug turned on in production!
+DEBUG = True
+
+ALLOWED_HOSTS = []
```
Sn1kls: Optional (good practice, not required for this task): `SECRET_KEY` and `DEBUG` are hardcoded here. It's fine for the task, but for a public repo it's better to read them from the environment:
```
import os
SECRET_KEY = os.environ["SECRET_KEY"]
DEBUG = os.environ.get("DEBUG", "False") == "True"
```

Keep the real values in a `.env` file, add `.env` to `.gitignore`, and commit a `.env.sample `with placeholders so others know what to set.

